PR #1, opened 2021-03-31
eventClass init

yuhao900914:


src/com/amplitude/Event.java
```diff
@@ -3,7 +3,104 @@
...
 public class Event {
+    // should move into Constants.java
+    public static final int MAX_PROPERTY_KEYS = 1024;
+    public static final int MAX_STRING_LENGTH = 1000;
+    public static final String TAG = "com.amplitude.Event"; //AmplitudeClient.class.getName();
 
-    private JsonObject eventProps;
+    private JSObject event;
+
+    public Event(String eventName, JSONObject eventProps, JSONObject groups, JSONObject groupProps,
```
dantetam, line 13: What is JSONObject groups and JSONObject groupProps? I feel like it's not needed for beta right now

src/com/amplitude/Event.java
```diff
@@ -3,7 +3,104 @@
...
+            this.event.put("groupProps", (groups == null) ? new JSONObject() : truncate(groups));
+
+            this.event.put("outOfSession", (outOfSession == true) ?  true : false);
+            this.event.put("timestamp", timestamp);
+
+            this.event.put("userProps",(userProps == null) ? new JSONObject() : truncate(userProps));
+            this.event.put("user_id", replaceWithJSONNull(user_id));
+            this.event.put("device_id", replaceWithJSONNull(device_id));
+            this.event.put("uuid", UUID.randomUUID().toString());
+            this.event.put("session_id", (this.outOfsession) ? -1 : session_id); // session_id = -1 if outOfSession = true;
```
dantetam, line 31: Need some solution to automatically set session id, that's unique to the user. Maybe a text file somewhere storing the session id number? How do the other SDKs do it? UNIX milliseconds time on start of session?

PR #3, opened 2021-04-12
microservice init

yuhao900914:


demo/src/com/amplitude/Demo.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+
+@SpringBootApplication
+public class Demo {
+
```
haoliu-amp, line 13: extra line

demo/src/com/amplitude/Demo.java
```diff
@@ -0,0 +1,33 @@
...
+	public static void main(String[] args) {
+		SpringApplication.run(Demo.class, args);
+	}
+
+	@Bean
+	public CommandLineRunner commandLineRunner(ApplicationContext ctx) {
+		return args -> {
+
+			System.out.println("Let's inspect the beans provided by Spring Boot:");
```
haoliu-amp, line 22: Could we change this to something meaningful?

demo/src/com/amplitude/Demo.java
```diff
@@ -0,0 +1,33 @@
...
+
+			String[] beanNames = ctx.getBeanDefinitionNames();
+			Arrays.sort(beanNames);
+			for (String beanName : beanNames) {
+				System.out.println(beanName);
+			}
+
```
haoliu-amp, line 29: extra line.

demo/src/com/amplitude/Demo.java
```diff
@@ -0,0 +1,33 @@
...
+		};
+	}
+
```
haoliu-amp, line 32: same

haoliu-amp (changes requested):
src/com/amplitude/Amplitude.java → app/src/com/amplitude/Amplitude.java 

This is not right. 
Should be something like:
main/java/com/amplitude/xxx.java
demo/java/com/amplitude/xxx.java
test/java/com/amplitude/xxx.java

demo/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+    module {
+        sourceDirs += file('src')
+    }
+}
+
+sourceSets.main.java.srcDirs = ['src']
+
+dependencies {
+    implementation 'org.json:json:20201115'
+    implementation 'org.springframework.boot:spring-boot-starter-web'
```
haoliu-amp, line 31: What's this package about? A package for starter?

demo/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+
+application {
+    mainClassName = "com.amplitude.Demo"
+}
+
+task runWithJavaExec(type: JavaExec) {
+    group = "Execution"
+    description = "Run the main class with JavaExecTask"
+    classpath = sourceSets.main.runtimeClasspath
+    main = "com.amplitude"
```
haoliu-amp, line 42: This is going to run the main method? Should this be `com.amplitude.Demo`? Also this is a micro service, why you bother running main?

haoliu-amp (changes requested):

demo/build.gradle
yuhao900914: That's the required dependency for `Spring Web`.  A starter for building web.

dantetam:
@yuhao900914 Did you test in IntelliJ how the two modules interact with each other? Are there any issues marking both `app` and `demo` both as source folders?

demo/build.gradle
dantetam: com.Amplitude.Demo was just for our internal testing of the API with `public static void main(String[] args)`

Ideally, yes, we would like a Gradle job like this one where the microservices app runs, sets up an endpoint that calls the Java SDK logEvent.

dantetam:
@haoliu-amp @yuhao900914 

Yes, the folder structure should be

src <---- Marked as src folder by gradle
      main
      demo
      test

src/demo/java/com/amplitude/DemoController.java
```diff
@@ -0,0 +1,12 @@
+package com.amplitude;
+
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RestController
+public class DemoController {
+	@RequestMapping("/")
+	public String index() {
+		return "Amplitude Java SDK Demo!";
```
haoliu-amp, line 10: Should try to log one event here.

haoliu-amp (changes requested):
Look much better!

haoliu-amp (approved):

dantetam (approved):

PR #4, opened 2021-04-17
Event class revamp

yuhao900914:


dantetam:
@yuhao900914 Do you know about this `com.fasterxml.jackson` library?

Is it entirely necessary and maintained?

yuhao900914:
> @yuhao900914 Do you know about this `com.fasterxml.jackson` library?
> 
> Is it entirely necessary and maintained?

There are two ways to convert the class to JSON, Gson, and Jackson. 
Gson is developed by Google and has more usage.  For others you might want to check [this](https://stackoverflow.com/questions/2378402/jackson-vs-gson). Looks like Jackson support more function and did better job in performancy. 
But the @JsonProperty annotation mentioned by Hao is the feature in Jackson. 

I think we can also convert by ourselves without using any third-partli brary. Like what we did in the current event class. Have a `getJsonObject()` function, create a JSONObject in it and return this JSONObject. So in this case, we just move all the code from the constructor to the  `getJsonObject()`  function. 

Any thought here? Which way should we choose?

src/main/java/com/amplitude/Event.java
```diff
@@ -7,104 +7,96 @@
...
-                value.substring(0, Constants.MAX_PROPERTY_KEYS);
-    }
-
-    public String toString() {
-        return this.event.toString();
-    }
-
-    public JSONObject getJsonObject() { return event; }
-
+    @JsonProperty("event_type")
```
haoliu-amp, line 15: Oh, if it needs extra library, then don't use it.

haoliu-amp: Just have a func `toJson` instead

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -50,8 +52,10 @@ public void logEvent(String name) {
 
     public void logEventWithProps(String eventName, JSONObject eventProps) {
         long time = System.currentTimeMillis();
-        Event event = new Event(eventName, eventProps, userProperties, "", Constants.SDK_VERSION,
-                lastEventId, sessionId, userId, deviceId, time);
+        Event event = new Event(eventName);
```
haoliu-amp, line 55: This is temporary, right?

src/main/java/com/amplitude/Event.java
```diff
@@ -7,104 +7,96 @@
...
+    public int event_id;
+    @JsonProperty("session_id")
+    public int session_id;
+    @JsonProperty("insert_id")
+    public int insert_id;
+
+    @JsonProperty("groups")
+    public JSONObject groups;
+
+    public Event(String _eventName) {
```
haoliu-amp, line 99: get rid of the underscore pls.

src/main/java/com/amplitude/Event.java
```diff
@@ -7,104 +7,96 @@
...
+    public int quantity;
+    @JsonProperty("revenue")
+    public double revenue;
+    @JsonProperty("productId")
+    public int productId;
+    @JsonProperty("revenueType")
+    public String revenueType;
+
+    @JsonProperty("event_id")
+    public int event_id;
```
haoliu-amp, line 90: Don't mix up hungarian & camel case together. Use camel case for all.

src/main/build.gradle
```diff
@@ -25,6 +25,8 @@ sourceSets.main.java.srcDirs = ['java']
 
 dependencies {
     implementation 'org.json:json:20201115'
+    compile group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: '2.8.8'
```
haoliu-amp, line 28: My bad, don't add extra dependencies.

haoliu-amp (changes requested):

src/main/java/com/amplitude/Amplitude.java
dantetam: I have already cleaned up this in the class. I'll fix the merge conflict when this is merged.

dantetam (approved):
Note: library can be set as a default value

src/main/java/com/amplitude/Event.java
```diff
@@ -10,34 +10,104 @@
 public class Event {
 
     public static final String TAG = "com.amplitude.Event"; //AmplitudeClient.class.getName();
+    private String eventType;
```
haoliu-amp, line 13: make this to be public too

src/main/java/com/amplitude/Event.java
```diff
@@ -10,34 +10,104 @@
 public class Event {
 
     public static final String TAG = "com.amplitude.Event"; //AmplitudeClient.class.getName();
```
haoliu-amp, line 12: why do you need this TAG

haoliu-amp (changes requested):

src/main/java/com/amplitude/Event.java
dantetam: We don't need the tag for now. Later on we'll add a few dedicated tagger (debug printer) so that people's apps won't be overridden with errors (for example, HTTP 400 response from Amplitude endpoint).

haoliu-amp: Then it should be Event.class.getName().

dantetam (approved):

PR #5, opened 2021-04-22
Backend fixes

dantetam:
The changes that we discussed to make a more generic server side API

### DIFF
--- a/src/demo/java/com/amplitude/DemoController.java
+++ b/src/demo/java/com/amplitude/DemoController.java
@@ -3,14 +3,13 @@ package com.amplitude;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.bind.annotation.RequestMapping;
 
-import com.amplitude.Amplitude;
 @RestController
 public class DemoController {
 	@RequestMapping("/")
 	public String index() {
 		Amplitude amplitude = Amplitude.getInstance("INSTANCE_NAME");
 		amplitude.init("8e07b9d451a7d07bd33f6e9ba5870f21");
-		amplitude.logEvent("Event name test Apr 14th 3:40 PM");
+		amplitude.logEvent(new Event("Test Event", "test_user_id"));
 		return "Amplitude Java SDK Demo: sending test event.";
 	}
 }
--- a/src/demo/java/com/amplitude/DemoController.java
+++ b/src/demo/java/com/amplitude/DemoController.java
@@ -3,6 +3,8 @@ package com.amplitude;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.bind.annotation.RequestMapping;
 
+import com.amplitude.Amplitude;
+
 @RestController
 public class DemoController {
 	@RequestMapping("/")
@@ -10,6 +12,7 @@ public class DemoController {
 		Amplitude amplitude = Amplitude.getInstance("INSTANCE_NAME");
 		amplitude.init("8e07b9d451a7d07bd33f6e9ba5870f21");
 		amplitude.logEvent(new Event("Test Event", "test_user_id"));
+		amplitude.setLogMode(AmplitudeLog.LogMode.DEBUG);
 		return "Amplitude Java SDK Demo: sending test event.";
 	}
 }
--- a/src/main/build.gradle
+++ b/src/main/build.gradle
@@ -27,4 +27,8 @@ dependencies {
     implementation 'org.json:json:20201115'
 }
 
+jar {
+    archiveBaseName = 'Amplitude-Java'
+    archiveVersion =  '1.0.0'
+}
 jar.baseName = 'Amplitude-Java'
--- a/src/main/build.gradle
+++ b/src/main/build.gradle
@@ -29,6 +29,5 @@ dependencies {
 
 jar {
     archiveBaseName = 'Amplitude-Java'
-    archiveVersion =  '1.0.0'
+    archiveVersion = '1.0.0'
 }
-jar.baseName = 'Amplitude-Java'
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -11,6 +11,8 @@ import java.util.concurrent.*;
 
 public class Amplitude {
 
+    public static final String TAG = Amplitude.class.getName();
+
     private static Map<String, Amplitude> instances = new HashMap<>();
     private String apiKey;
 
@@ -18,7 +20,7 @@ public class Amplitude {
 
     }
 
-    public static synchronized Amplitude getInstance(String instanceName) {
+    public static Amplitude getInstance(String instanceName) {
         if (!instances.containsKey(instanceName)) {
             Amplitude ampInstance = new Amplitude();
             instances.put(instanceName, ampInstance);
@@ -32,6 +34,12 @@ public class Amplitude {
 
     public void logEvent(String name) {
         logEvent(new Event(name));
+    }
+
+    public void logEvent(String eventName, JSONObject eventProps) {
+        Event event = new Event(eventName);
+        event.eventProperties = eventProps;
+        logEvent(event);
     }
 
     public void logEvent(Event event) {
@@ -56,26 +64,25 @@ public class Amplitude {
      */
     private Object syncHttpCall(Event event) {
         HttpsURLConnection connection = null;
+        InputStream inputStream = null;
         try {
             connection = (HttpsURLConnection) new URL(Constants.API_URL).openConnection();
             connection.setRequestMethod("POST");
             connection.setRequestProperty("Content-Type", "application/json");
             connection.setRequestProperty("Accept", "application/json");
-            //connection.setRequestProperty("Authorization", "Bearer " + apiKey);
             connection.setDoOutput(true);
 
             JSONObject bodyJson = new JSONObject();
             bodyJson.put("api_key", apiKey);
-            bodyJson.put("events", new JSONObject[]{event.getJsonObject()}); //event == null ? "[{}]" : event.toString());
+            bodyJson.put("events", event.toJsonObject());
 
             String bodyString = bodyJson.toString();
             OutputStream os = connection.getOutputStream();
             byte[] input = bodyString.getBytes("UTF-8");
             os.write(input, 0, input.length);
-            os.close();
+            //os.close();
 
             int responseCode = connection.getResponseCode();
-            InputStream inputStream;
             if (Constants.GOOD_RES_CODE_START <= responseCode &&
                     responseCode <= Constants.GOOD_RES_CODE_END) {
                 inputStream = connection.getInputStream();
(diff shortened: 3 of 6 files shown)


src/main/java/com/amplitude/Constants.java
```diff
@@ -2,11 +2,16 @@
...
+    long    NETWORK_TIMEOUT_MILLIS  = 10000;
+    String  SDK_LIBRARY             = "amplitude-java";
+    String  SDK_VERSION             = "1.0.0";
 
-    int MAX_PROPERTY_KEYS = 1024;
-    int MAX_STRING_LENGTH = 1000;
+    int     MAX_PROPERTY_KEYS       = 1024;
+    int     MAX_STRING_LENGTH       = 1000;
+
+    int     GOOD_RES_CODE_START     = 100;
```
haoliu-amp, line 13: Rename this to HTTP_STATUS_CONTINUE.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100

src/main/java/com/amplitude/Constants.java
```diff
@@ -2,11 +2,16 @@
...
+    int     GOOD_RES_CODE_END       = 399;
+    int     BAD_RES_CODE_START      = 400;
```
haoliu-amp, line 15: HTTP_STATUS_BAD_REQ

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -120,10 +78,11 @@ private Object syncHttpCall(Event event) {
             OutputStream os = connection.getOutputStream();
             byte[] input = bodyString.getBytes("UTF-8");
             os.write(input, 0, input.length);
+            //os.close();
 
             int responseCode = connection.getResponseCode();
-            InputStream inputStream;
-            if (100 <= responseCode && responseCode <= 399) {
+            if (Constants.GOOD_RES_CODE_START <= responseCode &&
+                    responseCode <= Constants.GOOD_RES_CODE_END) {
```
haoliu-amp, line 85: Why not `responseCode < 400`? You don't need 399 defined at all.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,12 +95,24 @@ private Object syncHttpCall(Event event) {
                 sb.append(output);
             }
 
-            if (responseCode >= 400) {
-                System.err.println("Warning, received error " + responseCode + " with message: " + sb.toString());
+            if (responseCode >= Constants.BAD_RES_CODE_START) {
+                AmplitudeLog.log(TAG, "Warning, received error HTTP code " + responseCode + " with message: " + sb.toString(),
+                        AmplitudeLog.LogMode.WARN);
+            }
+            else {
```
haoliu-amp, line 102: this should be in Ln 101

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,12 +95,24 @@ private Object syncHttpCall(Event event) {
...
+                AmplitudeLog.log(TAG, "Successful HTTP code " + responseCode + " with message: " + sb.toString(),
```
haoliu-amp, line 103: User can't control the logging info. What if they don't wanna see the debug info at all? 

You need to have a log level api exposed to the users to control the logging info.

src/main/java/com/amplitude/Event.java
```diff
@@ -41,7 +41,7 @@
 
     public String language;
     public String ip;
-    public String uuid;
+    public String uuid = UUID.randomUUID().toString();
```
haoliu-amp, line 44: Is this something we generate for customers? I don't see it [here](https://developers.amplitude.com/docs/http-api-v2)

src/main/java/com/amplitude/Event.java
```diff
@@ -90,7 +100,7 @@ public JSONObject toJsonObject() {
             event.put("android_id", replaceWithJSONNull(androidId));
             event.put("language", replaceWithJSONNull(language));
             event.put("ip", replaceWithJSONNull(ip));
-            event.put("uuid", UUID.randomUUID().toString());
+            event.put("uuid", replaceWithJSONNull(uuid));
```
haoliu-amp, line 103: Is this a legit property??

haoliu-amp (changes requested):

src/main/java/com/amplitude/Event.java
yuhao900914: I saw in android sdk, it assignS the uuid to UUID.randomUUID().toString();

src/demo/java/com/amplitude/DemoController.java
```diff
@@ -10,7 +10,7 @@
 	public String index() {
 		Amplitude amplitude = Amplitude.getInstance("INSTANCE_NAME");
 		amplitude.init("8e07b9d451a7d07bd33f6e9ba5870f21");
-		amplitude.logEvent("Test Event");
+		amplitude.logEvent("Event name test Apr 14th 3:40 PM");
```
yuhao900914, line 13: Minor thing but I think we should have a better name for the test logEvent.

src/main/build.gradle
```diff
@@ -27,4 +27,8 @@ dependencies {
     implementation 'org.json:json:20201115'
 }
 
+jar {
+    archiveBaseName = 'Amplitude-Java'
+    archiveVersion =  '1.0.0'
+}
 jar.baseName = 'Amplitude-Java'
```
yuhao900914, line 34: Is jar.baseName duplicate with jar {archiveBaseName XXX} ? 
I tested the jar.baseName =  'Amplitude-Java'. It will make the jar name to `Amplitude-Java`.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -44,18 +32,12 @@ public void init(String key) {
...
 
-    public void logEventWithProps(String eventName, JSONObject eventProps) {
-        long time = System.currentTimeMillis();
-        Event event = new Event(eventName);
-        event.eventProperties = eventProps;
-        event.deviceId = deviceId;
-        event.userId = userId;
-        lastEventId++;
-        logEvent(event);
+    public void logEvent(String name, String userId, String deviceId) {
```
yuhao900914, line 39: One question here. In Node.js, userId and deviceId are semi-required.  That should be either they provide userId or deviceId or have all of those is also ok. 
In this case,  did we handle that customers only have deviceId no userId?

src/main/java/com/amplitude/Event.java
```diff
@@ -123,7 +135,8 @@ protected JSONObject truncate(JSONObject object) {
         }
 
         if (object.length() > Constants.MAX_STRING_LENGTH) {
-            System.out.println("Warning: too many properties (more than 1000), ignoring");
+            AmplitudeLog.log(TAG, "Warning: too many properties (more than 1000), ignoring",
```
yuhao900914, line 138: Thanks for changing those to AmplitudeLog.

src/main/java/com/amplitude/Event.java
haoliu-amp: @yuhao900914 It might be legacy thing. Do we have uuid field in Node SDK?

src/demo/java/com/amplitude/DemoController.java
haoliu-amp: +1

src/main/java/com/amplitude/Event.java
yuhao900914: We has uuid field in Node SDK, it's an input field.

src/demo/java/com/amplitude/DemoController.java
dantetam: Yeah, oops, I'll fix it

src/main/java/com/amplitude/Amplitude.java
dantetam: They should use `logEvent(name, null, deviceId)`

src/main/java/com/amplitude/Amplitude.java
dantetam: Will do so. I don't think HTTP codes are changing any time soon

src/main/java/com/amplitude/Amplitude.java
haoliu-amp: I think they might not need this func at all. Let's just keep `logEvent(Event e)`

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    public enum LogMode {
+        DEBUG(1),
+        WARN(2),
+        ERROR(3),
+        OFF(4);
+
+        private int index;
```
haoliu-amp, line 32: Let's call this  `level`

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static void setLogMode(LogMode logMode) {
+        AmplitudeLog.LOG_MODE = logMode;
+    }
+
+    public static void log(String tag, String message) {
+        log(tag, message, LogMode.DEBUG);
+    }
+
+    public static void log(String tag, String message, LogMode messageMode) {
```
haoliu-amp, line 15: You can have other functions to call this with LogMode set.
`debug(...)`
`error(...)`
`warn(...)`

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -0,0 +1,38 @@
+package com.amplitude;
+
+public class AmplitudeLog {
+
+    private static LogMode LOG_MODE = LogMode.DEBUG;
+
+    public static void setLogMode(LogMode logMode) {
```
haoliu-amp, line 7: We should give user to set this when they create Amplitude class, which means we need to have an API for user in `Amplitude` to set the log level.

I don't see this is being called anywhere btw.

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -0,0 +1,38 @@
+package com.amplitude;
+
+public class AmplitudeLog {
+
+    private static LogMode LOG_MODE = LogMode.DEBUG;
```
haoliu-amp, line 5: Default should be OFF or ERROR

haoliu-amp (changes requested):

src/main/build.gradle
dantetam: Good catch. jar.baseName is deprecated. This does the same function. I just found documentation [here](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html).

src/demo/java/com/amplitude/DemoController.java
```diff
@@ -3,14 +3,13 @@
 import org.springframework.web.bind.annotation.RestController;
```
haoliu-amp, line 3: Why the import is gone?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,12 +87,23 @@ private Object syncHttpCall(Event event) {
                 sb.append(output);
             }
 
-            if (responseCode >= 400) {
-                System.err.println("Warning, received error " + responseCode + " with message: " + sb.toString());
+            if (responseCode >= Constants.HTTP_STATUS_BAD_REQ) {
```
haoliu-amp, line 90: Combine the logic with the previous if/else

haoliu-amp (changes requested):

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -0,0 +1,46 @@
...
+    public void error(String tag, String message) {
+        log(tag, message, LogMode.ERROR);
+    }
+
+    public void log(String tag, String message, LogMode messageMode) {
+        if (messageMode.level >= logMode.level) {
+            if (messageMode.level >= LogMode.ERROR.level) {
+                System.err.println(tag + ": " + message);
+            }
+            else {
```
haoliu-amp, line 28: nit: move this to Ln27

haoliu-amp (approved):

Issue #6, opened 2021-05-10
Incompatible class type version 59 with JDK 8


bhadkenaresh841:
Unable to use amplitude java SDK on JDK 1.8 as Amplitude 1.0.1 is compiled on JDK 16, Throwing runtime exception class file has wrong version 59.0, should be 52.0.

dantetam:
Hello @bhadkenaresh841 ,

My apologies. I'm currently resolving the issue right now. This library should be compiled with Java 1.8.

Dante

dantetam:
Hello @bhadkenaresh841 , this should be fixed in version 1.1.0 of the Java SDK.

bhadkenaresh841:
@dantetam Thank you for the quick resolution.

PR #7, opened 2021-05-12
Clean up project specs and compile for Java 8 target

dantetam:
The library should be compiled for Java 8 as it does not have any special features beyond Java 8, and should be compiled properly on projects that still use Java 8.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,4 +1,4 @@
-rootProject.name = 'Amplitude-Java-SDK'
+rootProject.name = 'amplitude-java-sdk'
 include 'main'
 project(':main').projectDir = file('src/main')
 include 'demo'
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,4 +1,4 @@
-rootProject.name = 'Amplitude Java SDK'
+rootProject.name = 'Amplitude-Java-SDK'
 include 'main'
 project(':main').projectDir = file('src/main')
 include 'demo'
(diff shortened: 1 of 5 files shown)


settings.gradle
```diff
@@ -1,5 +1,5 @@
-rootProject.name = 'Amplitude Java SDK'
+rootProject.name = 'Amplitude-Java-SDK'
```
haoliu-amp, line 1: Can we make it all lower case?

src/main/java/com/amplitude/Constants.java
```diff
@@ -5,7 +5,7 @@
     String  API_URL                 = "https://api2.amplitude.com/2/httpapi";
     long    NETWORK_TIMEOUT_MILLIS  = 10000;
     String  SDK_LIBRARY             = "amplitude-java";
-    String  SDK_VERSION             = "1.0.0";
+    String  SDK_VERSION             = "1.1.0";
```
haoliu-amp, line 8: You can auto generate this just like Android.

haoliu-amp (changes requested):

src/main/java/com/amplitude/Constants.java
dantetam: BuildConfig is specific to Android runtime. I can't figure out a clean way to do it now. 

I saw one approach in Java where the Java file reads the Gradle file in:
```
Properties properties = new Properties()
properties.load(project.rootProject.file('local.properties').newDataInputStream())
def sdkDir = properties.getProperty('sdk.dir')
```

haoliu-amp: https://stackoverflow.com/questions/46805846/how-do-you-specify-buildconfigfield-in-gradle-java-library-project-build-script We can do it later.

haoliu-amp (approved):

dantetam:
Fixes https://github.com/amplitude/Amplitude-Java/issues/6

PR #11, opened 2021-06-03
Throttling

yuhao900914:
Create Response class move syncHttpCallWithEventsBuffer function in to Response class.
Implement the retry logic.


src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
...
+            }
+        } else if (onceReponse.status == Status.SUCCESS) {
+            shouldRetry = false;
+        }
+        return retryEventsOnceResult.getResult(shouldRetry, shouldReduceEventCount, eventIndicesToRemove);
+    }
+
+    private static void retryEventsOnLoop(String userId, String deviceId, String apiKey) {
+        List<Event> eventsBuffer = getRetryBuffer(userId, deviceId);
+        int[] eventCountHolder = new int[]{eventsBuffer.size()};
```
yuhao900914, line 138: Because in Java 8 we must use a final variable in the lambda function. 
But in our case, the events count need to be modified afterward. So this is a workaround.
Any thought on this part?

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private static String getStringValueWithKey(JSONObject json, String key) {
+        return json.has(key) && json.getString(key) != null ? json.getString(key) : "";
+    }
+
+    private static JSONObject getJSONObjectValueWithKey(JSONObject json, String key) {
+        return (json.has(key) && !json.isNull(key)) ? json.getJSONObject(key) : new JSONObject();
+    }
+
+    protected static Status getCodeStatus(int code) {
```
qingzhuozhen, line 152: Should we consider moving this to `Status` class?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
...
+        result.shouldRetry = shouldRetry;
+        result.shouldReduceEventCount = shouldReduceEventCount;
+        result.eventIndicesToRemove = eventIndicesToRemove;
+        return result;
+    }
+}
+
+class Retry {
+    // Use map to record the events are currently in retry queue.
+    private static Map<String, Map<String, List<Event>>> idToBuffer = new HashMap<String, Map<String, List<Event>>>();
```
qingzhuozhen, line 27: Would this be thread safe?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
+package com.amplitude;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+class retryEventsOnceResult {
```
qingzhuozhen, line 10: For class name maybe we can name it `RetryEventsOnceResult`

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
...
+        List<Event> eventsToRetry = deviceToBufferMap.get(deviceId);
+        if (eventsToRetry != null && eventsToRetry.size() == 0) {
+            deviceToBufferMap.remove(deviceId);
+        }
+        if (deviceToBufferMap.size() == 0) {
+            idToBuffer.remove(userId);
+        }
+    }
+
+    private static int[] JSONArrayToIntArray(JSONArray jsonArray) {
```
qingzhuozhen, line 70: I see this logic is also used in `convertJSONArrayToIntArray`, should we just create a helper class and move these method there? and the naming could be camelCase

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
...
+        Thread retryThread =
+                new Thread(() -> {
+                    for (int numRetries = 0; numRetries < retryTimes; numRetries++) {
+                        long sleepDuration = Constants.RETRY_TIMEOUTS[numRetries];
+                        if ( eventCountHolder[0] <= 0 || eventCountHolder[0] > eventsBuffer.size())
+                            break;
+                        try {
+                            Thread.sleep(sleepDuration);
+                            boolean isLastTry = numRetries == Constants.RETRY_TIMEOUTS.length - 1;
+                            List<Event> eventsToRetrys = eventsBuffer.subList(0, eventCountHolder[0]);
```
qingzhuozhen, line 153: nit: typo, maybe just `ventsToRetry`

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -88,83 +76,18 @@ public synchronized void flushEvents() {
             List<Event> eventsInTransit = new ArrayList<>(eventsToSend);
             eventsToSend.clear();
             CompletableFuture.supplyAsync(() -> {
-                int responseCode = syncHttpCallWithEventsBuffer(eventsInTransit);
-                if (responseCode >= Constants.HTTP_STATUS_MIN_RETRY && responseCode <= Constants.HTTP_STATUS_MAX_RETRY) {
-                    eventsToSend.addAll(eventsInTransit);
-                    tryToFlushEventsIfNotFlushing();
+                Response response = Response.syncHttpCallWithEventsBuffer(eventsInTransit, apiKey);
+                int responseCode = response.code;
+                Status status = Response.getCodeStatus(responseCode);
```
qingzhuozhen, line 81: this could just be `Status status = response.status`? I guess we always populate the status together with code.

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,173 @@
...
+        return res;
+    }
+
+    /*
+     * Use HTTPUrlConnection object to make async HTTP request,
+     * using data from event like device, class name, event props, etc.
+     *
+     * @return The response code
+     */
+    protected static Response syncHttpCallWithEventsBuffer(List<Event> events, String apiKey) {
```
haoliu-amp, line 67: This function is to make a request. Why do we put it inside Response class?

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,173 @@
...
+public class Response {
+    protected int code;
+    protected Status status;
+    protected String error;
+    protected JSONObject SuccessBody;
+    protected JSONObject InvalidRequestBody;
+    protected JSONObject RateLimitBody;
+    protected static Response res;
+
+    private static Response createNewResponseFromJSON(JSONObject json) {
```
haoliu-amp, line 20: Can rename it to `populateResponse`.

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,173 @@
...
+            JSONArray jsonArray = json.getJSONArray(key);
+            int[] intArray = new int[jsonArray.length()];
+            for (int i = 0; i < jsonArray.length(); i++) {
+                intArray[i] = jsonArray.getInt(i);
+            }
+            return intArray;
+        }
+    }
+
+    private static String getStringValueWithKey(JSONObject json, String key) {
```
haoliu-amp, line 144: Is it worth to put it in it's own Util class? Same qq for next function.

haoliu-amp (changes requested):

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,253 @@
...
+            arr[i] = jsonArray.optInt(i);
+        }
+        return arr;
+    }
+
+    private static int[] collectInvalidEventIndices(Response response) {
+        List<Integer> invalidIndices = new ArrayList<Integer>();
+        if (response.status == Status.INVALID && response.InvalidRequestBody != null) {
+            JSONObject eventsWithInvalidFields = response.InvalidRequestBody.getJSONObject("eventsWithInvalidFields");
+            JSONObject eventsWithMissingFields = response.InvalidRequestBody.getJSONObject("eventsWithMissingFields");
```
qingzhuozhen, line 82: Seems the logic for both Fields processing is the same, we could extract it to a method and reuse

src/main/java/com/amplitude/Retry.java
qingzhuozhen: Good catch, one thought for this is could we just create the variable and assign it to another variable inside lambda then modify the new one? like
```
int eventsBufferSize = eventsBuffer.size()
....
() -> {
    int eventCount = eventsBufferSize;
    ...
    eventCount /= 2;

}
``` So we don't modify the local variables inside lambda. Other thought is some other workaround like AtomicInteger

File: src/main/java/com/amplitude/Amplitude.java [hunk 12]
```diff
@@ -88,83 +76,18 @@ public synchronized void flushEvents() {
...
+                if (status == Status.INVALID ||
+                    status == Status.PAYLOAD_TOO_LARGE ||
```
qingzhuozhen, line 83: nit: could be a helper method like `shouldRetryForStatus`, to isolate the status check a bit

src/main/java/com/amplitude/Response.java
yuhao900914: @dantetam  suggested that I should move the HTTP codes into here from Response.java. 
Should we create another class like HttpCall and move this in?

qingzhuozhen: Ideally Response class just handle the part like parsing, and make request calls could be in a separate class, and in future if we support more calls we can also add in that class.

src/main/java/com/amplitude/Retry.java
dantetam: Alyssa looked up some special code on stack overflow. For some reason this doesn't work? I guess the way it is here in Qingzhuo's code, `eventsBufferSize` is effectively final.

src/main/java/com/amplitude/Utils.java
```diff
@@ -0,0 +1,31 @@
...
+        return json.has(key) && json.getString(key) != null ? json.getString(key) : "";
+    }
+
+    public static JSONObject getJSONObjectValueWithKey(JSONObject json, String key) {
+        return (json.has(key) && !json.isNull(key)) ? json.getJSONObject(key) : new JSONObject();
+    }
+
+    public static int[] jsonArrayToIntArray(JSONArray jsonArray) {
+        int[] intArray = new int[jsonArray.length()];
+        for (int i = 0; i < intArray.length; ++i) {
```
dantetam, line 17: i++ is the standard form, but it doesn't matter here.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,244 @@
...
+                        .collect(Collectors.toList());
+            }
+        } else if (response.status == Status.SUCCESS) {
+            return;
+        }
+
+        for (Event event : eventsToRetry) {
+            String userId = (event.userId != null) ? event.userId : "";
+            String deviceId = (event.deviceId != null) ? event.deviceId : "";
+            if ((userId.length() > 0) || (deviceId.length() > 0)) {
```
dantetam, line 215: Unnecessary parentheses inside the if statement

dantetam: Symbols are evaluated in a special and set (i.e. static and working with static types) way. Very fun stuff, this is just extra reading: https://introcs.cs.princeton.edu/java/11precedence/

src/main/java/com/amplitude/Retry.java
dantetam: int[] is an object. Does object vs primitive change the ability to use inside of a lambda function?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+class RetryEventsOnceResult {
+    protected boolean shouldRetry;
+    protected boolean shouldReduceEventCount;
+    protected int[] eventIndicesToRemove;
+    protected static RetryEventsOnceResult result;
+
+    protected static RetryEventsOnceResult getResult(boolean shouldRetry, boolean shouldReduceEventCount, int[] eventIndicesToRemove) {
```
dantetam, line 17: Is there a reason this is not just a standard constructor?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,244 @@
...
+                    shouldRetry = false;
+                }
+            }
+            // Reduce the payload to reduce risk of throttling
+            shouldReduceEventCount = true;
+        } else if (onceReponse.status == Status.PAYLOAD_TOO_LARGE) {
+            shouldRetry = true;
+        } else if (onceReponse.status == Status.INVALID) {
+            if (events.size() == 1) {
+                shouldRetry = false;
```
dantetam, line 119: Can remove the extra `shouldRetry = false`. By default, it starts as false.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,244 @@
...
+        int[] invalidIndices = new int[]{};
+        if (response.status == Status.INVALID && response.InvalidRequestBody != null) {
+            List<Integer> invalidFieldsIndices = collectIndicesWithRequestBody(response.InvalidRequestBody, "eventsWithInvalidFields");
+            List<Integer> missingFieldsIndices = collectIndicesWithRequestBody(response.InvalidRequestBody, "eventsWithMissingFields");
+            invalidFieldsIndices.addAll(missingFieldsIndices);
+            Collections.sort(invalidFieldsIndices);
+            int[] allInvalidEventIndices = invalidFieldsIndices.stream().mapToInt(i -> i).toArray();
+            return allInvalidEventIndices;
+        }
+        return invalidIndices;
```
dantetam, line 96: Can be shortened to `return new int[]{}`
In Java, this syntax is fine:
```
int[] arr = new int[5];
arr = new int[]{};

int[] arr2 = {};
int[] arr3 = {5, 10};
```

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+public class Response {
+    protected int code;
+    protected Status status;
+    protected String error;
+    protected JSONObject SuccessBody;
+    protected JSONObject InvalidRequestBody;
+    protected JSONObject RateLimitBody;
+    protected static Response res;
```
dantetam, line 18: I'm sorry, why is this declared here? This should only be initialized once, right? `populateResponse` should only touch this variable once, based on some JSON?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -1,18 +1,7 @@
...
-
-import javax.net.ssl.HttpsURLConnection;
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.OutputStream;
-import java.net.URL;
-
+import java.net.HttpCookie;
```
qingzhuozhen, line 3: `HttpCookie` could be removed, I guess this is by ide..

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -0,0 +1,80 @@
+package com.amplitude;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import javax.net.ssl.HttpsURLConnection;
+import java.io.*;
```
qingzhuozhen, line 7: Is `java.io.*` generated automatically? I wonder if we should just import the necessary one rather than using the `*` and import all under io.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,244 @@
+package com.amplitude;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.util.*;
```
qingzhuozhen, line 6: same concern for `util.*`

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,60 @@
+package com.amplitude;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import javax.net.ssl.HttpsURLConnection;
```
qingzhuozhen, line 6: some imports are not needed any more

src/main/java/com/amplitude/Response.java
yuhao900914: You are right, it didn't been used outside. Will move into populateResponse.

src/main/java/com/amplitude/Retry.java
yuhao900914: @dantetam  It works with the object like what I did now. Just want to ask if there is any concern or if there is any better way to do it. I will also talked about this with Qingzhuo today.

src/main/java/com/amplitude/HttpCall.java
dantetam: It is, IntelliJ does it.

yuhao900914: Yes, it's generated automatically. Just thinking that might import lots of things we don't need.
I removed the auto-generated import and import the things we need.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.HashMap;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Collections;
+
+class RetryEventsOnceResult {
+    protected boolean shouldRetry;
+    protected boolean shouldReduceEventCount;
+    protected int[] eventIndicesToRemove;
```
qingzhuozhen, line 19: nit: let's leave a line below..

qingzhuozhen:
Couple thoughts:
1. could we consider using format tool or plugin to maintain the format and align future development? This could also save some effort on clean up the imports. Doesn't have to be pre-commit hooks now, but maybe something we can align on our side.
2. should we add some testing to make sure this is thread safe? If it is too big, we can do it in separate ticket.
Talked a bit with @dantetam about this, would like to know others' thoughts as well.

src/main/java/com/amplitude/Retry.java
```diff
@@ -151,30 +147,28 @@ private static void retryEventsOnLoop(String userId, String deviceId, String api
...
-                                eventCountHolder[0] = eventCountHolder[0] / 2;
+                                eventCount /= 2;
                             }
-                            // Clean up the events
-                            eventsBuffer.subList(0, eventCountHolder[0]).clear();
-                            eventsInRetry -= eventCountHolder[0];
                         } catch (InterruptedException e) {
                         }
                     }
+                    eventsInRetry -= eventCount;
```
yuhao900914, line 171: Since the logic in node js in here is cleaning up the eventsBuffer after the retry loop for all cases, so we don't need to put the eventsBuffer back to the idToBuffer map. 
For reference: https://github.com/amplitude/Amplitude-Node/blob/1f6d6578aa060ecd53f867721924b75b2ac22e1c/packages/node/src/retry/defaultRetry.ts#L272

src/main/java/com/amplitude/Retry.java
```diff
@@ -56,15 +56,11 @@ protected RetryEventsOnceResult(boolean shouldRetry, boolean shouldReduceEventCo
     }
 
     // Cleans up the id in the buffer map if the job is done
-    private static void cleanUpBuffer(String userId, String deviceId) {
```
yuhao900914: Update the cleanUpBuffer function like this, since it is only called after we remove the eventsBuffer from the idToBuffer map in retryEventsOnLoop.

src/main/java/com/amplitude/Retry.java
dantetam: +1, Alyssa and I discussed this. It appears all the events are being removed anyway.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,252 @@
...
+    this.shouldReduceEventCount = shouldReduceEventCount;
+    this.eventIndicesToRemove = eventIndicesToRemove;
+  }
+}
+
+class Retry {
+  // Use map to record the events are currently in retry queue.
+  private static Map<String, Map<String, List<Event>>> idToBuffer =
+      new ConcurrentHashMap<String, Map<String, List<Event>>>();
+  private static int eventsInRetry = 0;
```
qingzhuozhen, line 32: I guess `eventsInRetry` is not thread safe here, and it can be accessed by multiple threads, should we use `AtomicInteger` instead?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,254 @@
...
+                    int numEventsRemoved = 0;
+                    for (int i = 0; i < eventIndicesToRemove.length; i++) {
+                      int index = eventIndicesToRemove[i];
+                      if (index < eventCount) {
+                        eventsBuffer.remove(i);
+                        numEventsRemoved += 1;
+                      }
+                    }
+                    eventCount -= numEventsRemoved;
+                    eventsInRetry.set(eventsInRetry.intValue() - numEventsRemoved);
```
qingzhuozhen, line 167: when we update eventsInRetry, could we use method like `decrementAndGet` to guarantee the thread safety? Two atomic operation together won't guarantee the atomic though.

yuhao900914: decrementAndGet only decrements by one. But I didn't find a way to decrease a certain number except using .set() method.

qingzhuozhen: I see, could we use addAndGet with negative value? or maybe updateAndGet?

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,268 @@
...
+        Map<String, List<Event>> deviceToBufferMap = idToBuffer.get(userId);
+        if (deviceToBufferMap == null) {
+          deviceToBufferMap = new ConcurrentHashMap<String, List<Event>>();
+          idToBuffer.put(userId, deviceToBufferMap);
+        }
+        List<Event> retryBuffer = deviceToBufferMap.get(deviceId);
+        if (retryBuffer == null) {
+          retryBuffer = new ArrayList<Event>();
+          deviceToBufferMap.put(deviceId, retryBuffer);
+          retryEventsOnLoop(userId, deviceId, apiKey);
```
qingzhuozhen, line 240: I guess we don't need this line any more

yuhao900914: yeah, my bad. Will remove this.

src/main/java/com/amplitude/Retry.java
```diff
@@ -0,0 +1,268 @@
...
+        }
+        eventsInRetry.incrementAndGet();
+        retryBuffer.add(event);
+      }
+    }
+    idToBuffer.forEach(
```
qingzhuozhen, line 246: Curious why this is preferred comparing to store the userId and deviceId information in a list then looping it? And I guess ConcurrentHashMap helps to avoid ConcurrentModificationException here? If we have to have two thread in onEventsError, would that kick off two threads for same user and deviceId or run before one loop in one thread ends? I guess if the events happen to share the user and device id we may hit this case as well, however if they don't would this just happens to create that case more?

src/main/java/com/amplitude/Response.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.json.JSONObject;
+
+public class Response {
+  protected int code;
+  protected Status status;
+  protected String error;
+  protected JSONObject SuccessBody;
+  protected JSONObject InvalidRequestBody;
+  protected JSONObject RateLimitBody;
```
qingzhuozhen, line 11: nit: ideally this is camelCase

qingzhuozhen (approved):
lgtm now, in future when we add tests we should try to cover more tests case including mutlithreads cases to make sure we don't miss anything.

PR #14, opened 2021-06-14
Add Javadocs for documentation of all public facing methods and class variables

dantetam:
Will do the release right after this is approved


src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,162 @@
...
+   */
   public long timestamp = System.currentTimeMillis();
+  /** The current Latitude of the user. */
   public double locationLat;
+  /** The current Longitude of the user. */
   public double locationLng;
 
+  /** The current version of your application. */
   public String appVersion;
+
```
qingzhuozhen, line 38: missing comments for `versionName`?

src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,162 @@
 
 public class Event {
 
+  /** A unique identifier for your event. */
   public String eventType;
```
yuhao900914, line 13: We should have another empty line after each variable define.

src/main/java/com/amplitude/Event.java
dantetam: I couldn't actually find documentation for this variable? @yuhao900914 ?

src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,162 @@
...
+   * use the IP address to reverse lookup a user's location (city, country, region, and DMA).
+   * Amplitude has the ability to drop the location and IP address from events once it reaches our
+   * servers. You can submit a request to our platform specialist team here to configure this for
+   * you.
+   */
   public String ip;
+  /**
+   * A dictionary of key-value pairs that represent additional data to be sent along with the event.
+   * You can store property values in an array. Date values are transformed into string values.
+   * Object depth may not exceed 40 layers.
```
yuhao900914, line 84: Feels like we don't have the `40 layers check` in the event class. Should we also add that logic? or that should in a separate ticket.

src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,162 @@
...
   public String idfv;
+  /** (Android) Google Play Services advertising ID */
   public String adid;
+  /** (Android) Android ID (not the advertising ID) */
   public String androidId;
-
+  /** The (human) language set by the user. */
   public String language;
+  /**
+   * The IP address of the user. Use "$remote" to use the IP address on the upload request. We will
```
yuhao900914, line 74: What does `Use "$remote" to use the IP address on the upload request` mean?

src/main/java/com/amplitude/Event.java
yuhao900914: I did add the version_name in the nodejs even type. But I cannot find where it comes from now. Will check.

src/main/java/com/amplitude/Event.java
qingzhuozhen: For the format part, maybe @dantetam can also run the google formatter on local to reflect? I guess that can help the formats.

src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,162 @@
...
 
+  /**
+   * The price of the item purchased. Required for revenue data if the revenue field is not sent.
+   * You can use negative values to indicate refunds.
+   */
   public double price;
+  /** The quantity of the item purchased. Defaults to 1 if not specified. */
   public int quantity;
+  /**
+   * revenue = price quantity. If you send all 3 fields of price, quantity, and revenue, then (price
```
qingzhuozhen, line 102: `price * quantity`? Something miss in the middle?

src/main/java/com/amplitude/Event.java
dantetam: This check is part of the HTTP V2 API, looks like

src/main/java/com/amplitude/Event.java
dantetam: From HTTP v2 API. This means, pass "$remote" into this field to use the IP address of the HTTP request, as opposed to a custom set IP.

src/main/java/com/amplitude/Event.java
dantetam: Good catch. Appears to be a typo on the HTTP V2 API as well. Will alert the team

src/main/java/com/amplitude/Event.java
dantetam: Ran the formatter. No spaces needed from them, but I decided to add the spaces for better readability.

Later on, customers will use the Javadocs to figure out which fields of `Event` to set.

src/main/java/com/amplitude/Event.java
yuhao900914: Just discuss this with @kelvin-lu, since this also exists in node js sdk. The version name does not exist in current http v2 documents, so that's might be the legacy code.

src/main/java/com/amplitude/Event.java
```diff
@@ -9,56 +9,189 @@
...
+   */
   public String insertId = UUID.randomUUID().toString();
 
+  /**
+   * This feature is only available to Enterprise customers who have purchased the Accounts add-on.
+   * This field adds a dictionary of key-value pairs that represent groups of users to the event as
+   * an event-level group. You can only track up to 5 groups.
+   */
   public JSONObject groups;
+
```
qingzhuozhen, line 169: nit: do we need docs for `groupProperties`?

yuhao900914 (approved):
LGTM;

qingzhuozhen (approved):
lgtm

PR #18, opened 2021-06-17
fix type for productId and sessionId

yuhao900914:


dantetam (approved):
LGTM! Checked both fields in HTTP API V2 reference

PR #21, opened 2021-06-22
implement batch http call

yuhao900914:


src/main/java/com/amplitude/Constants.java
```diff
@@ -3,6 +3,13 @@
 public interface Constants {
 
   String API_URL = "https://api2.amplitude.com/2/httpapi";
+  String BATCH_API_URL = "https://api2.amplitude.com/batch";
+
+  enum httpCallMode {
```
qingzhuozhen, line 8: let's not put enum inside Constant class... if you want you can possibly leave in in this file, but move out of the Constant class. Also, you can just use the new enum class to manage the value of url.
also the enum class name should be `HttpCallMode`

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -59,6 +61,13 @@ public void init(String key) {
     apiKey = key;
   }
 
+  public void isBatchMode(Boolean isBatchMode) {
```
qingzhuozhen, line 64: the method name is confusing though, by `isBatchMode` it sounds to be a method checking if it is batch mode, other than trying to set to use batch.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -108,11 +117,17 @@ public synchronized void flushEvents() {
       eventsToSend.clear();
       CompletableFuture.supplyAsync(
           () -> {
-            Response response = HttpCall.syncHttpCallWithEventsBuffer(eventsInTransit, apiKey);
+            Response response = null;
+            if (httpCallMode == Constants.httpCallMode.REGULAR_HTTPCALL) {
+              response =
+                  new GeneralHttpCall(eventsInTransit, apiKey).syncHttpCallWithEventsBuffer();
```
qingzhuozhen, line 123: I believe there is no need to initiate a new object for this each time...

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -108,11 +117,17 @@ public synchronized void flushEvents() {
...
+            } else {
+              response = new BatchHttpCall(eventsInTransit, apiKey).syncHttpCallWithEventsBuffer();
+            }
             Status status = response.status;
 
             if (Retry.shouldRetryForStatus(status)) {
-              Retry.sendEventsWithRetry(eventsInTransit, apiKey, response);
+              Retry.sendEventsWithRetry(eventsInTransit, apiKey, response, httpCallMode);
```
qingzhuozhen, line 130: we don't need to send apiKey in this case, right? Since the new object already have it

src/main/java/com/amplitude/Amplitude.java
qingzhuozhen: not a fan of put if else in here.. can we have a private method `getHttpCall`? Wondering if we should have a private variable to hold it.

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,30 @@
...
    * @return The response object which contains a code and other information
    */
-  protected static Response syncHttpCallWithEventsBuffer(List<Event> events, String apiKey) {
+  public String apiKey;
+  public List<Event> events;
+  public String apiUrl;
+
+  protected HttpCall(List<Event> events, String apiKey, String apiUrl) {
+    this.apiKey = apiKey;
+    this.events = events;
```
qingzhuozhen, line 30: Do we need to have the events in constructor? I don't feel we do in this case..
Can we just have apiKey, apiUrl? apiKey is better to be the enum type.
Then the method syncHttpCallWithEventsBuffer will take a list of event as argument.

src/main/java/com/amplitude/Amplitude.java
yuhao900914: in Retry.retryEventsOnce, we have another HTTP call. So I think we still need to pass the apiKey?

qingzhuozhen: yeah, that's for you are passing the `httpCallMode`, instead of `HttpCall` right? And you need to construct the `httpCall` later, but if you don't need the events in `HttpCall`, do you still need to pass them to construct a new `HttpCall` in Retry?

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,35 @@
...
+  public String apiKey;
+  public String apiUrl;
+
+  protected HttpCall(String apiKey, String apiUrl) {
+    this.apiKey = apiKey;
+    this.apiUrl = apiUrl;
+  }
+
+  protected Response syncHttpCallWithEventsBuffer(List<Event> events) {
+    System.out.println("*********");
```
qingzhuozhen, line 38: clean up the console logs

src/main/java/com/amplitude/Retry.java
```diff
@@ -102,8 +102,8 @@ private static void cleanUpBuffer(String userId) {
   }
 
   private static RetryEventsOnceResult retryEventsOnce(
-      String userId, String deviceId, List<Event> events, String apiKey) {
-    Response onceReponse = HttpCall.syncHttpCallWithEventsBuffer(events, apiKey);
+      String userId, String deviceId, List<Event> events,HttpCall httpCall) {
```
qingzhuozhen, line 105: space in the argument, could run the formatter tool later.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -57,6 +61,17 @@ public static Amplitude getInstance(String instanceName) {
    */
   public void init(String key) {
     apiKey = key;
+    httpCall = new GeneralHttpCall(apiKey);
+  }
+
+  public void setHttpCallMode(Boolean isBatchMode) {
```
qingzhuozhen, line 67: maybe add some docs for this method, would `useBatchMode` wording a bit better?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -57,6 +61,17 @@ public static Amplitude getInstance(String instanceName) {
...
+    if (isBatchMode) {
+      httpCallMode = HttpCallMode.BATCH_HTTPCALL;
+      httpCall = new BatchHttpCall(apiKey);
+    } else {
+      httpCallMode = HttpCallMode.REGULAR_HTTPCALL;
+      httpCall = new GeneralHttpCall(apiKey);
```
qingzhuozhen, line 73: one thing I feel is the httpCall here doesn't really rely on the httpCallMode, and if I remove the httpCallMode, it won't really affect or break anything... like in init, it just initiate on with genera but not really caring about the httpCallMode. One thing we can consider is have a method createHttpCall or updateHttpCall, with a param HttpCallMode, inside it checks if there is update to call mode, then update httpCall as necessary, and use that as the only method that can update httpCall. So we don't have multiple methods or place to update httpCall.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -98,6 +113,10 @@ private void tryToFlushEventsIfNotFlushing() {
     }
   }
 
+  private Response getHttpCall(List<Event> eventsInTransit) {
```
qingzhuozhen, line 116: I see, originally I was suggesting to use `getHttpCall` to return the instance of `HttpCall`, if you want to make a call in this method, then maybe give a different name

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,35 @@
...
    * Use HTTPUrlConnection object to make async HTTP request,
    * using data from event like device, class name, event props, etc.
    *
    * @return The response object which contains a code and other information
    */
-  protected static Response syncHttpCallWithEventsBuffer(List<Event> events, String apiKey) {
+  public String apiKey;
+  public String apiUrl;
+
+  protected HttpCall(String apiKey, String apiUrl) {
```
qingzhuozhen, line 32: one possible further way we can consider is in HttpCallMode, each enum will hold the url value, then HttpCall just take HttpCallMode as constructor. This way it can force the HttpCall to only use limited url. As long as HttpCallMode enum is not exposed to public that is fine for me.

qingzhuozhen: Another thing we can consider is this constructor just take an apiKey, then in abstract class we just have an abstract method getApiUrl, which returns a string. Then we don't need to think about the last comment. And in each sub class, we can just overwrite the method getApiUrl.

src/main/java/com/amplitude/Amplitude.java
yuhao900914: Yea, I realize the same thing. You are right we should check if the current HttpCallMode is changed or not. If it's not changed, we don't need to do anything.

src/main/java/com/amplitude/HttpCall.java
yuhao900914: I Will choose the second suggestion since we can take advantage of the abstract class and make the constructor simpler.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -99,20 +129,21 @@ private void tryToFlushEventsIfNotFlushing() {
...
    */
   public synchronized void flushEvents() {
     if (eventsToSend.size() > 0) {
       List<Event> eventsInTransit = new ArrayList<>(eventsToSend);
       eventsToSend.clear();
       CompletableFuture.supplyAsync(
           () -> {
-            Response response = HttpCall.syncHttpCallWithEventsBuffer(eventsInTransit, apiKey);
+            Response response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+            ;
```
qingzhuozhen, line 142: clean up extra `;`

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,35 @@
...
+public abstract class HttpCall {
   /*
    * Use HTTPUrlConnection object to make async HTTP request,
    * using data from event like device, class name, event props, etc.
    *
    * @return The response object which contains a code and other information
    */
-  protected static Response syncHttpCallWithEventsBuffer(List<Event> events, String apiKey) {
+  public String apiKey;
+  public String apiUrl;
```
qingzhuozhen, line 30: no need to keep this `apiUrl` in class

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,35 @@
 
 import java.util.List;
 
-public class HttpCall {
+enum HttpCallMode {
+  REGULAR_HTTPCALL,
+  BATCH_HTTPCALL
+};
```
qingzhuozhen, line 20: no need for `;`

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,21 +14,35 @@
...
 
+public abstract class HttpCall {
   /*
    * Use HTTPUrlConnection object to make async HTTP request,
    * using data from event like device, class name, event props, etc.
    *
    * @return The response object which contains a code and other information
    */
-  protected static Response syncHttpCallWithEventsBuffer(List<Event> events, String apiKey) {
+  public String apiKey;
```
qingzhuozhen, line 29: `apiKey` doesn't ned to be public

src/main/java/com/amplitude/BatchHttpCall.java
```diff
@@ -0,0 +1,14 @@
+package com.amplitude;
+
+public class BatchHttpCall extends HttpCall {
+  private static String apiUrl = Constants.BATCH_API_URL;
+
+  protected BatchHttpCall(String apiKey) {
```
qingzhuozhen, line 6: Nit: The constructor here is not necessary now, may be ok to keep it though

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -51,16 +57,39 @@ public static Amplitude getInstance(String instanceName) {
...
+  public void useBatchMode(Boolean isBatchMode) {
+    updateHttpCall(isBatchMode ? HttpCallMode.BATCH_HTTPCALL : HttpCallMode.REGULAR_HTTPCALL);
+  }
+
+  private void updateHttpCall(HttpCallMode updatedHttpCallMode) {
+    if (updatedHttpCallMode != httpCallMode) {
+      if (updatedHttpCallMode == HttpCallMode.BATCH_HTTPCALL) {
+        httpCall = new BatchHttpCall(apiKey);
+      } else if (updatedHttpCallMode == HttpCallMode.REGULAR_HTTPCALL) {
+        httpCall = new GeneralHttpCall(apiKey);
```
qingzhuozhen, line 85: I believe you need to update the httpCallMode after.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -51,16 +57,39 @@ public static Amplitude getInstance(String instanceName) {
...
-   * Set the API key for this instance of Amplitude. API key is necessary to authorize
-   * and route events to the current Amplitude project.
+   * Set the API key for this instance of Amplitude. API key is necessary to authorize and route
+   * events to the current Amplitude project.
+   *
    * @param key the API key from Amplitude website
    */
   public void init(String key) {
     apiKey = key;
+    httpCall = new GeneralHttpCall(apiKey);
```
qingzhuozhen, line 67: could use the new method as well, just add a null check in the new method

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -51,16 +56,42 @@ public static Amplitude getInstance(String instanceName) {
...
+   * HTTP V2 Batch API.
+   *
+   * @param isBatchMode if using batch upload or not;
+   */
+  public void useBatchMode(Boolean isBatchMode) {
+    updateHttpCall(isBatchMode ? HttpCallMode.BATCH_HTTPCALL : HttpCallMode.REGULAR_HTTPCALL);
+  }
+
+  private void updateHttpCall(HttpCallMode updatedHttpCallMode) {
+    if (httpCallMode == null) {
```
qingzhuozhen, line 80: oh, originally I was thinking about httpCall null check, as you define httpCallMode in constructor. If you do it this way, it is ok too, just you can skip the null check, or combine them together.

`if (httpCallMode == null || htppCallMode != updatedHttpCallMode)`

qingzhuozhen (approved):
lgtm now. I believe you also tested and verified as well.

PR #22, opened 2021-06-24
fix: add customize AmplitudeInvalidAPIKeyException add throw this exception when input invalid apiKey

yuhao900914:


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -81,8 +83,8 @@ private void updateHttpCall(HttpCallMode updatedHttpCallMode) {
       httpCallMode = updatedHttpCallMode;
       httpCall =
           (updatedHttpCallMode == HttpCallMode.BATCH_HTTPCALL)
-              ? new BatchHttpCall(apiKey)
-              : new GeneralHttpCall(apiKey);
+              ? new BatchHttpCall(apiKey, logger)
```
qingzhuozhen, line 86: Not a fan to pass the logger with it, normal way to get a logger is like `Logger LOGGER = LogManager.getLogger();`, we can see if we want to adjust `AmplitudeLog` to do similar like getInstance

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -137,7 +139,11 @@ public synchronized void flushEvents() {
       eventsToSend.clear();
       CompletableFuture.supplyAsync(
           () -> {
-            Response response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+            Response response = null;
+            try {
+              response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+            } catch (AmplitudeInvalidAPIKeyException e) {
+            }
```
dantetam, line 146: I think in general, the flow should be to stop the whole Amplitude SDK on any invalid API key. Thoughts?

dantetam: User was specifically asking, and I agree, for a hard noticeable stop in the program when the API key is wrong or missing.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -137,7 +139,11 @@ public synchronized void flushEvents() {
       eventsToSend.clear();
       CompletableFuture.supplyAsync(
           () -> {
-            Response response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+            Response response = null;
+            try {
+              response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+            } catch (AmplitudeInvalidAPIKeyException e) {
```
qingzhuozhen, line 145: I see we are suppressing by catching in here, so for user who accidentally input in invalid apiKey, how would they know if the apiKey is invalid?

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -79,6 +82,8 @@ protected Response syncHttpCallWithEventsBuffer(List<Event> events) {
       }
       JSONObject responseJson = new JSONObject(sb.toString());
       responseBody = Response.populateResponse(responseJson);
+    } catch(AmplitudeInvalidAPIKeyException e) {
```
qingzhuozhen, line 85: if we are catching the exception already, this method is not throwing any more exception I guess

src/main/java/com/amplitude/Response.java
```diff
@@ -10,9 +12,14 @@
   protected JSONObject invalidRequestBody;
   protected JSONObject rateLimitBody;
 
-  protected static Response populateResponse(JSONObject json) {
+  protected static Response populateResponse(JSONObject json) throws AmplitudeInvalidAPIKeyException {
     Response res = new Response();
     int code = json.getInt("code");
+    String invalidAPIKeyError = "Invalid API key: .*";
+    String errorMsg = Utils.getStringValueWithKey(json, "error");
+    if (errorMsg.matches(invalidAPIKeyError)) {
```
qingzhuozhen, line 20: the check could be in a private method

src/main/java/com/amplitude/Response.java
```diff
@@ -10,9 +12,14 @@
   protected JSONObject invalidRequestBody;
   protected JSONObject rateLimitBody;
 
-  protected static Response populateResponse(JSONObject json) {
+  protected static Response populateResponse(JSONObject json) throws AmplitudeInvalidAPIKeyException {
     Response res = new Response();
     int code = json.getInt("code");
+    String invalidAPIKeyError = "Invalid API key: .*";
```
qingzhuozhen, line 18: I guess this was tested, and maybe the check should be under the certain status block?

src/main/java/com/amplitude/exception/AmplitudeInvalidAPIKeyException.java
```diff
@@ -0,0 +1,30 @@
+package com.amplitude.exception;
+
+public class AmplitudeInvalidAPIKeyException extends Exception {
+
+    private final int code;
```
qingzhuozhen, line 5: why we need code here?

src/main/java/com/amplitude/exception/AmplitudeInvalidAPIKeyException.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public AmplitudeInvalidAPIKeyException(int code) {
+        super();
+        this.code = code;
+    }
+
+    public AmplitudeInvalidAPIKeyException(String message, Throwable cause, int code) {
```
qingzhuozhen, line 12: I guess we don't need to have constructors that we are using

src/main/java/com/amplitude/Retry.java
```diff
@@ -176,7 +178,7 @@ private static void retryEventsOnLoop(String userId, String deviceId, HttpCall h
                   if (shouldReduceEventCount && !isLastTry) {
                     eventCount /= 2;
                   }
-                } catch (InterruptedException e) {
+                } catch (InterruptedException | AmplitudeInvalidAPIKeyException e) {
```
qingzhuozhen, line 181: Same thing I guess we are not really getting this exception here. Also by catching the exception in there, would the purpose is we keep trying.

src/main/java/com/amplitude/Retry.java
```diff
@@ -102,7 +104,7 @@ private static void cleanUpBuffer(String userId) {
   }
 
   private static RetryEventsOnceResult retryEventsOnce(
-      String userId, String deviceId, List<Event> events, HttpCall httpCall) {
+      String userId, String deviceId, List<Event> events, HttpCall httpCall) throws AmplitudeInvalidAPIKeyException {
```
qingzhuozhen, line 107: if we catch inside, then this is misleading to throws exception

src/main/java/com/amplitude/exception/AmplitudeInvalidAPIKeyException.java
```diff
@@ -0,0 +1,30 @@
+package com.amplitude.exception;
+
+public class AmplitudeInvalidAPIKeyException extends Exception {
```
qingzhuozhen, line 3: could we have an `AmplitudeException` then let this extends `AmplitudeException`

src/main/java/com/amplitude/HttpCall.java
yuhao900914: Actually, the exception will throw in here. And the catch in Amplitude.java is not doing anything but required surrounded the try-catch block from ItelliJ.

src/main/java/com/amplitude/exception/AmplitudeInvalidAPIKeyException.java
yuhao900914: I went over some customize examples and tried to make our customize exception covered all the cases. I also think if we do need these constructors or not since we only use one of it. 
Thanks for your suggestion will only leave one constructor.public AmplitudeInvalidAPIKeyException(String message).

src/main/java/com/amplitude/Amplitude.java
qingzhuozhen: yeah, ideally the flow should stop and we are not sending anything further.

src/main/java/com/amplitude/HttpCall.java
qingzhuozhen: the exception is thrown in `Response.populateResponse` right? And as it is get caught here, this method is not throwing any more `AmplitudeInvalidAPIKeyException`, the Intellij only gives hint since `throws AmplitudeInvalidAPIKeyException` is in signature. So if you want to throw again at here, you may want to add like `throw e` right? I guess if we want to support hard stop, then we don't need to log for this.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -1,5 +1,8 @@
 package com.amplitude;
 
+import com.amplitude.exception.AmplitudeInvalidAPIKeyException;
+
+import java.rmi.ServerException;
```
qingzhuozhen, line 5: This is not used?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -135,16 +138,27 @@ public synchronized void flushEvents() {
...
-            }
-            return null;
-          });
+      CompletableFuture<Response> asyncHandler =
+          CompletableFuture.supplyAsync(
+              () -> {
+                Response response = null;
+                try {
+                  response = httpCall.syncHttpCallWithEventsBuffer(eventsInTransit);
+                } catch (AmplitudeInvalidAPIKeyException e) {
```
qingzhuozhen, line 147: why catch and print here?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -135,16 +138,27 @@ public synchronized void flushEvents() {
...
+                  e.printStackTrace();
+                }
+                return response;
+              });
+
+      try {
+        Response response = asyncHandler.join();
```
qingzhuozhen, line 154: Do we really need to break this up?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -135,16 +138,27 @@ public synchronized void flushEvents() {
...
+        Status status = response.status;
+        if (Retry.shouldRetryForStatus(status)) {
+          Retry.sendEventsWithRetry(eventsInTransit, response, httpCall);
+        }
+      } catch (CompletionException e) {
+        e.printStackTrace();
```
qingzhuozhen, line 160: I mean, ideally we may want this method `flushEvents` just throw the Invalid api key exception to let user aware. And also, we don't really want to print stack trace in our code.

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -94,7 +97,7 @@ protected Response syncHttpCallWithEventsBuffer(List<Event> events) {
 
         }
       }
-      return responseBody;
     }
+    return responseBody;
```
qingzhuozhen, line 101: I see, so in this case, if the exception is thrown and not caught, nothing will be returned.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,15 +138,27 @@ public synchronized void flushEvents() {
...
+              response -> {
+                Status status = response.status;
+                if (Retry.shouldRetryForStatus(status)) {
+                  Retry.sendEventsWithRetry(eventsInTransit, response, httpCall);
+                }
+              })
+          .exceptionally(
+              exception -> {
+                logger.error("Invalid API Key", exception.getMessage());
+                return null;
```
qingzhuozhen, line 160: Would we consider throw the exception here? I guess for now user wouldn't get the exception?

src/main/java/com/amplitude/Status.java
```diff
@@ -20,6 +20,14 @@
   /** environment error.. E.g. disconnected from network */
   SYSTEM_ERROR;
 
+  protected static Boolean hasInvalidAPIKey(String errorMsg){
```
qingzhuozhen, line 23: I guess this method could be in `Response` class, since this method doesn't really have anything related to status

src/main/java/com/amplitude/exception/AmplitudeException.java
```diff
@@ -0,0 +1,7 @@
+package com.amplitude.exception;
+
+public class AmplitudeException extends Exception {
+    public AmplitudeException(String message) {
+        super(message);
+    }
+}
```
qingzhuozhen, line 7: new line, I guess we can run formatter later to cover this. Or you can config the plugin to run on every save

src/main/java/com/amplitude/Retry.java
```diff
@@ -177,6 +180,11 @@ private static void retryEventsOnLoop(String userId, String deviceId, HttpCall h
                     eventCount /= 2;
                   }
                 } catch (InterruptedException e) {
+                } catch (AmplitudeInvalidAPIKeyException e) {
+                  // The retry logic should only be executed after the API key checking passed.
+                  // This catch AmplitudeInvalidAPIKeyException is just for handling
+                  // retryEventsOnce in thread.
+                  Thread.currentThread().interrupt();
```
qingzhuozhen, line 187: looks fine, I guess return or interrupt both work. Do we also need this line for `InterruptedException`?

src/main/java/com/amplitude/exception/AmplitudeException.java
yuhao900914: Sounds good, will check how to make the formatter run for every save, that's much more convenient.

src/main/java/com/amplitude/Retry.java
yuhao900914: Should we interrupt the thread or throw RuntimeException?
https://stackoverflow.com/a/22621931/11969222

qingzhuozhen: `AmplitudeInvalidAPIKeyException` is already an exception here, normally if we don't catch it should just stop thread from continue

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,15 +138,27 @@ public synchronized void flushEvents() {
...
-            Status status = response.status;
-
-            if (Retry.shouldRetryForStatus(status)) {
-              Retry.sendEventsWithRetry(eventsInTransit, response, httpCall);
-            }
-            return null;
-          });
+              () -> {
+                Response response = null;
+                try {
```
qingzhuozhen, line 143: Nit: indent

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -136,15 +138,27 @@ public synchronized void flushEvents() {
...
+          .thenAcceptAsync(
+              response -> {
+                Status status = response.status;
+                if (Retry.shouldRetryForStatus(status)) {
+                  Retry.sendEventsWithRetry(eventsInTransit, response, httpCall);
+                }
+              })
+          .exceptionally(
+              exception -> {
+                logger.error("Invalid API Key", exception.getMessage());
```
qingzhuozhen, line 159: Let's provide the feedback and in doc tells we will log exceptions

yuhao900914: Sure will do after merge

src/main/java/com/amplitude/Response.java
```diff
@@ -10,7 +12,16 @@
   protected JSONObject invalidRequestBody;
   protected JSONObject rateLimitBody;
 
-  protected static Response populateResponse(JSONObject json) {
+  protected static Boolean hasInvalidAPIKey(String errorMsg) {
```
qingzhuozhen, line 15: nit: could be `private` and `boolean`?

src/main/java/com/amplitude/Response.java
```diff
@@ -22,8 +33,9 @@ protected static Response populateResponse(JSONObject json) {
       res.successBody.put("payloadSizeBytes", json.getInt("payload_size_bytes"));
       res.successBody.put("serverUploadTime", json.getLong("server_upload_time"));
     } else if (status == Status.INVALID) {
-      res.invalidRequestBody = new JSONObject();
       res.error = Utils.getStringValueWithKey(json, "error");
+      if (hasInvalidAPIKey(res.error)) throw new AmplitudeInvalidAPIKeyException(res.error);
```
qingzhuozhen, line 37: nit: let's make it two lines

src/main/java/com/amplitude/Response.java
```diff
@@ -10,7 +12,16 @@
   protected JSONObject invalidRequestBody;
   protected JSONObject rateLimitBody;
 
-  protected static Response populateResponse(JSONObject json) {
+  protected static Boolean hasInvalidAPIKey(String errorMsg) {
+    String invalidAPIKeyError = "Invalid API key: .*";
+    if (errorMsg.matches(invalidAPIKeyError)) {
```
qingzhuozhen, line 17: this could be just `return errorMsg.matches(invalidAPIKeyError)` or 
```
if (errorMsg.matches(invalidAPIKeyError)) {
throws ...
}
```

src/main/java/com/amplitude/Response.java
yuhao900914: My bad.

src/main/java/com/amplitude/Amplitude.java
yuhao900914: Will not fix this since it's the result after running google java format.

src/main/java/com/amplitude/Response.java
yuhao900914: Will not fix this since it's the result after running google java format.

qingzhuozhen (approved):

PR #25, opened 2021-07-06
test: add unit tests

qingzhuozhen:
### Summary

add unit tests

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-iOS/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -1,13 +1,15 @@
 package com.amplitude;
 
-import com.amplitude.exception.AmplitudeInvalidAPIKeyException;
-
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
 import java.util.Queue;
-import java.util.concurrent.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import com.amplitude.exception.AmplitudeInvalidAPIKeyException;
 
 public class Amplitude {
 
--- a/src/test/java/com/amplitude/AmplitudeLogTest.java
+++ b/src/test/java/com/amplitude/AmplitudeLogTest.java
@@ -1,0 +1,71 @@
+package com.amplitude;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class AmplitudeLogTest {
+  private ByteArrayOutputStream outContent;
+  private ByteArrayOutputStream errContent;
+  private final PrintStream originalOut = System.out;
+  private final PrintStream originalErr = System.err;
+  private final AmplitudeLog amplitudeLog = new AmplitudeLog();
+
+  @BeforeEach
+  public void setUpStreams() {
+    outContent = new ByteArrayOutputStream();
+    errContent = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(outContent));
+    System.setErr(new PrintStream(errContent));
+  }
+
+  @AfterEach
+  public void restoreStreams() throws IOException {
+    outContent.close();
+    errContent.close();
+    System.setOut(originalOut);
+    System.setErr(originalErr);
+  }
+
+  @ParameterizedTest
+  @MethodSource("logArguments")
+  public void testLog(
+      AmplitudeLog.LogMode logMode,
+      String expectedErrorLog,
+      String expectedWarnLog,
+      String expectedDebugLog) {
+    amplitudeLog.setLogMode(logMode);
+    amplitudeLog.error("Test", "error message");
+    assertEquals(expectedErrorLog, errContent.toString().trim());
+    amplitudeLog.warn("Test", "warn message");
+    assertEquals(expectedWarnLog, outContent.toString().trim());
+    amplitudeLog.log("Test", "debug message");
+    assertEquals(expectedDebugLog, outContent.toString().trim());
+  }
+
+  static Stream<Arguments> logArguments() {
+    return Stream.of(
+        arguments(AmplitudeLog.LogMode.ERROR, "Test: error message", "", ""),
+        arguments(
+            AmplitudeLog.LogMode.WARN,
+            "Test: error message",
+            "Test: warn message",
+            "Test: warn message"),
+        arguments(
+            AmplitudeLog.LogMode.DEBUG,
+            "Test: error message",
+            "Test: warn message",
+            "Test: warn message\nTest: debug message"),
+        arguments(AmplitudeLog.LogMode.OFF, "", "", ""));
+  }
+}
--- a/src/test/java/com/amplitude/AmplitudeTest.java
+++ b/src/test/java/com/amplitude/AmplitudeTest.java
@@ -1,10 +1,33 @@
 package com.amplitude;
 
-import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.assertNotSame;
 import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import com.amplitude.exception.AmplitudeInvalidAPIKeyException;
+import com.amplitude.util.EventsGenerator;
+
+@ExtendWith(MockitoExtension.class)
 public class AmplitudeTest {
+
+  private final String apiKey = "test-apiKey";
 
   @Test
   public void testGetInstance() {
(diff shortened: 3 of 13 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -1,13 +1,15 @@
...
 import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
 import java.util.Queue;
-import java.util.concurrent.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import com.amplitude.exception.AmplitudeInvalidAPIKeyException;
```
dantetam, line 12: Yeah, moving forward, this is a good idea Qingzhuo. Not having the IDE (IntelliJ) automatically use asterisk imports. SDKs should be lightweight.

src/test/java/com/amplitude/AmplitudeLogTest.java
```diff
@@ -0,0 +1,71 @@
...
+    System.setErr(originalErr);
+  }
+
+  @ParameterizedTest
+  @MethodSource("logArguments")
+  public void testLog(
+      AmplitudeLog.LogMode logMode,
+      String expectedErrorLog,
+      String expectedWarnLog,
+      String expectedDebugLog) {
```
dantetam, line 46: I like the framework here for testing outputs to System.err/out, this makes sense for testing all the modes

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -17,4 +40,99 @@ public void testGetInstance() {
...
+    Amplitude amplitude = Amplitude.getInstance("test");
+    amplitude.init(apiKey);
+    amplitude.useBatchMode(useBatch);
+    amplitude.setLogMode(AmplitudeLog.LogMode.OFF);
+    List<Event> events = EventsGenerator.generateEvents(10, 5, 6);
+    HttpCall httpCall = getMockHttpCall(amplitude, useBatch);
+    Response response = new Response();
+    response.code = 200;
+    response.status = Status.SUCCESS;
+    CountDownLatch latch = new CountDownLatch(1);
```
dantetam, line 58: I'm sure this is right, but please explain to me. Why are we waiting when the response is mocked?

src/test/java/com/amplitude/EventTest.java
```diff
@@ -0,0 +1,45 @@
...
+          new Event("test event", null, null);
+        },
+        "Event must have one defined userId and/or deviceId");
+  }
+
+  @Test
+  public void testToJsonObject() {
+    Event event = new Event("test event", "test-user");
+    String longStr =
+        "This is a long string that is too long for an event, and going to be truncated later when we send the event.";
```
dantetam, line 25: Might have to revisit this one. The test is correct, however, there will be different behavior for batch API vs normal API. I believe the Batch API only starts truncating at much longer strings. Also maybe use code that more easily shows the length of this string.

src/test/java/com/amplitude/util/MockURLStreamHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class MockURLStreamHandler extends URLStreamHandler implements URLStreamHandlerFactory {
+
+  private final Map<URL, URLConnection> connections = new HashMap<>();
+
+  private final static MockURLStreamHandler instance = new MockURLStreamHandler();
+
+  public static MockURLStreamHandler getInstance(){
```
dantetam, line 17: nit: the space (){

src/test/java/com/amplitude/AmplitudeTest.java
qingzhuozhen: yeah, the issue here is for our call is inside the async or another thread, so even if response is called, there is no way to guarantee our assertion called after the response is returned and all the async part finished. So in this case we can also do `Thread.sleep` in tests, but the waiting time getting tricky, a long time is fine but getting the tests to always taking a longer time, a short time could make the tests flakey, which is not ideal. So just use the CountDownLatch here to await and make sure our assertion happens after the call to prevent the sleep part.

src/test/java/com/amplitude/util/MockURLStreamHandler.java
qingzhuozhen: oh, let me run the formatter again on this.

src/test/java/com/amplitude/util/EventsGenerator.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.UUID;
+
+import com.amplitude.Event;
+
+public class EventsGenerator {
+  public static List<Event> generateEvents(int eventCount) {
+    return generateEvents(eventCount, 1, 1);
+  }
+
+  public static List<Event> generateEvents(int eventCount, int userCount, int deviceCount) {
```
dantetam, line 15: Rename these parameters to `userIdLength` and `deviceIdLength`? A bit confusing to me as I thought multiple users were generated as some sort of event/user/group property.

src/test/java/com/amplitude/EventTest.java
qingzhuozhen: Right now I don't think there is any logic in event is concerning if it is batch api or normal api though. And I thought the difference between batch and normal apis is more on how many events we can send with api.

I guess I can make this shorter then assert the length before the truncate runs, which should be easier to see the difference.

src/test/java/com/amplitude/util/EventsGenerator.java
qingzhuozhen: how about `userIdCount` or `deviceIdCount`, `userIdLength` or `deviceIdLength` is confusing me to consider it is about the length of user id or device id. The goal for this helper method is to just generate events for multiple users or devices.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -17,4 +40,99 @@ public void testGetInstance() {
...
+    verify(httpCall, times(1)).syncHttpCallWithEventsBuffer(anyList());
+  }
+
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
+  public void testLogEventWithInvalidKeyException(boolean useBatch)
+      throws InterruptedException, NoSuchFieldException, IllegalAccessException,
+          AmplitudeInvalidAPIKeyException {
+    Amplitude amplitude = Amplitude.getInstance("test");
+    amplitude.init(apiKey);
```
yuhao900914, line 78: Curious about this part, looks like all functions for testing the logEevent shared the same fake apiKey, why they will return different answer after call the `syncHttpCallWithEventsBuffer` ?

src/test/java/com/amplitude/EventTest.java
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "Event must have one defined userId and/or deviceId");
+  }
+
+  @Test
+  public void testToJsonObject() {
+    Event event = new Event("test event", "test-user");
+    String longStr = "Long string to be truncated.";
+    for (int i = 0; i < 7; i++) {
+      longStr += longStr;
```
yuhao900914, line 26: Since this is the test, might be we don't need to use StringBuilder instead for improving the efficiency?

src/test/java/com/amplitude/EventTest.java
```diff
@@ -0,0 +1,45 @@
+package com.amplitude;
```
yuhao900914, line 1: For testing the event, do we need to test we can get the right value from the event object after initiating a new Event?

src/test/java/com/amplitude/EventTest.java
```diff
@@ -0,0 +1,45 @@
...
+    assertEquals(3584, longStr.length());
+    JSONObject eventProperties = new JSONObject();
+    eventProperties.put("event_type", "test event type");
+    JSONArray eventMsgArray = new JSONArray();
+    eventMsgArray.put(longStr);
+    eventMsgArray.put(longStr);
+    eventProperties.put("event_message", eventMsgArray);
+    event.eventProperties = eventProperties;
+
+    JSONObject truncatedEvent = event.toJsonObject();
```
yuhao900914, line 37: Do we need to cover all possible keys in Event Class, and check if we can get the correct value from `event.toJsonObject()`?

src/test/java/com/amplitude/AmplitudeTest.java
dantetam: Ohhh, that makes a lot of sense. That's really deep async logic. Didn't actually remember this despite the fact I did work on the async calls of Java SDK. This `CountDownLatch` is probably the cleanest way to handle this.

src/test/java/com/amplitude/AmplitudeTest.java
dantetam: The responses are mocked. The api key is just to complete the object?

src/test/java/com/amplitude/EventTest.java
dantetam: +1 should use stringBuilder when building long strings, that's a good catch Alyssa. As an example, the Android tests are extremely long. Takes around two to three minutes to finish

src/test/java/com/amplitude/EventTest.java
dantetam: I don't mind just skipping it. @qingzhuozhen if you really, really wanted to test every single param. Otherwise, I'm just going to assume the json library has its own tests.

dantetam (changes requested):
In general, I like all the tests. All the framework, the logic, and the actual assertions being tested. Just left a few concerns in my comments

src/test/java/com/amplitude/util/EventsGenerator.java
dantetam: Yeah, userCount to me sounds like a number of unique user IDs or something.

src/test/java/com/amplitude/EventTest.java
dantetam: Yeah, that's the issue I was getting at. The HTTP V2 API vs Batch API treats this differently. Now that we support batch API, does this complicate the tests, and does it require new code behavior?

So the normal HTTP V2 API doesn't allow strings longer than 1024 characters, in its documentation. I don't see a limit in Batch API, only a file size of 20MB per payload. I swear I saw a different limit for Batch API somewhere.

qingzhuozhen: Yeah, for now the event logic and truncate logic is separate with the api though, and we are relying on the api response now to tell. If we have a know document about this, then we can decide if we want to adjust the truncate logic based on the api type. For the test now, we can leave for it now

src/test/java/com/amplitude/util/EventsGenerator.java
qingzhuozhen: yeah, that's actually the goal for this, generate event for various users.

src/test/java/com/amplitude/AmplitudeTest.java
qingzhuozhen: The api key here doesn't really matter, just leave it like how we normally use the library. Also we need the api key to complete the object and http call initiation.

src/test/java/com/amplitude/EventTest.java
qingzhuozhen: Hmm, for the test here, I guess the efficiency improvement is trivial though.. It doesn't hurt to change it, just won't be that much difference for limited loop.

src/test/java/com/amplitude/EventTest.java
qingzhuozhen: I guess there is no need to set or assert all possible keys here, I add some assertion on the ones we are having values. Later we can cover more if needed.

yuhao900914: Sounds good.

src/test/java/com/amplitude/EventTest.java
dantetam: These strings are getting fairly long right, for the later truncate string test cases. Thanks for adding stringBuilder

src/test/java/com/amplitude/HttpCallTest.java
```diff
@@ -0,0 +1,252 @@
...
+  @MethodSource("httpCallArguments")
+  public void testHttpCallWithIOException(HttpCallMode httpCallMode, URL url)
+      throws IOException, AmplitudeInvalidAPIKeyException {
+    HttpsURLConnection connection = mock(HttpsURLConnection.class);
+    when(connection.getOutputStream()).thenThrow(new IOException("test IOException thrown"));
+    mockURLStreamHandler.setConnection(url, connection);
+    HttpCall httpCall = getHttpCallFromCallMode(httpCallMode);
+    List<Event> events = EventsGenerator.generateEvents(1);
+    Response response = httpCall.syncHttpCallWithEventsBuffer(events);
+    assertEquals(0, response.code);
```
yuhao900914, line 227: Since we change to give the TIMEOUT status when throwing IOException, should we also changed here? 
https://github.com/amplitude/Amplitude-Java/pull/26

qingzhuozhen: yeah, good catch, let me pull the changes, I guess that change will break this test

qingzhuozhen: already updated

dantetam (approved):
Looks good to me! We've resolved every issue I've brought up. Is there anything else @yuhao900914 ?

yuhao900914 (approved):
LGTM, please run all tests to make sure nothing will break.  Approved this PR.

PR #26, opened 2021-07-14
fix: Status codes minor issues

dantetam:
### Summary

Fixes [AMP-37687](https://amplitude.atlassian.net/browse/AMP-37687)

### Checklist

* [ X ] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-iOS/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?

### DIFF
--- a/src/main/java/com/amplitude/HttpCall.java
+++ b/src/main/java/com/amplitude/HttpCall.java
@@ -87,7 +87,7 @@ public abstract class HttpCall {
       // Also SocketTimeoutException, when the HTTP request times out.
       JSONObject timesOutResponse = new JSONObject();
       timesOutResponse.put("status", Status.TIMEOUT);
-      timesOutResponse.put("code", 0);
+      timesOutResponse.put("code", 408);
       responseBody = Response.populateResponse(timesOutResponse);
     } finally {
       if (inputStream != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/amplitude/HttpCall.java
```diff
@@ -87,7 +87,7 @@ protected Response syncHttpCallWithEventsBuffer(List<Event> events)
       // Also SocketTimeoutException, when the HTTP request times out.
       JSONObject timesOutResponse = new JSONObject();
       timesOutResponse.put("status", Status.TIMEOUT);
```
yuhao900914, line 89: Should this be `Status.UNKNOWN` ?

dantetam: I've decided to use the TIMEOUT status for all of these errors/cases. In the documentation, it says that TIMEOUT is used for any server side or client side HTTP request that times out. This is the best way to organize it I think, let me know if you had more context or ideas

yuhao900914: Yea, but I looked at the ticket in Jira, it says the test case will return UNKNOW in the response.

dantetam: Yeah, that's why the number was changed to 408. So it should point to TIMEOUT

yuhao900914: Oh, I see.

yuhao900914 (approved):

qingzhuozhen (approved):

PR #30, opened 2021-09-14
fix: add revenue props in the event payload when there has revenue or…

yuhao900914:
… quantity and price info

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Event.java
+++ b/src/main/java/com/amplitude/Event.java
@@ -115,7 +115,7 @@ public class Event {
    * The price of the item purchased. Required for revenue data if the revenue field is not sent.
    * You can use negative values to indicate refunds.
    */
-  public double price;
+  public Double price;
 
   /** The quantity of the item purchased. Defaults to 1 if not specified. */
   public int quantity;
@@ -125,7 +125,7 @@ public class Event {
    * (price * quantity) will be used as the revenue value. You can use negative values to indicate
    * refunds.
    */
-  public double revenue;
+  public Double revenue;
 
   /**
    * An identifier for the item purchased. You must send a price and quantity or revenue with this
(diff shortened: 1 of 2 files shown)


src/main/java/com/amplitude/Event.java
```diff
@@ -240,11 +240,18 @@ public JSONObject toJsonObject() {
...
-      event.put("productId", productId);
-      event.put("revenueType", revenueType);
+
+      boolean shouldLogRevenueProps = (revenue != null || (price != null && quantity > 0));
+      if (shouldLogRevenueProps) {
+        event.put("price", price);
+        event.put("quantity", quantity);
+        event.put(
+            "revenue",
+            (revenue != null && price != null && quantity > 0) ? quantity * price : revenue);
```
qingzhuozhen, line 250: I am a bit confused here, so if there exists revenue and price and quantity then we do our calculation?

src/main/java/com/amplitude/Event.java
```diff
@@ -240,11 +240,18 @@ public JSONObject toJsonObject() {
       event.put(
           "user_properties",
           (userProperties == null) ? new JSONObject() : truncate(userProperties));
-      event.put("price", price);
-      event.put("quantity", quantity);
-      event.put("revenue", revenue);
-      event.put("productId", productId);
-      event.put("revenueType", revenueType);
+
+      boolean shouldLogRevenueProps = (revenue != null || (price != null && quantity > 0));
```
qingzhuozhen, line 244: Could we add a simple test case for this?

src/main/java/com/amplitude/Event.java
yuhao900914: it’s how api v2 says.

https://developers.amplitude.com/docs/http-api-v2

revenue = price * quantity. If you send all 3 fields of price, quantity, and revenue, then (price * quantity) will be used as the revenue value. You can use negative values to indicate refunds.

yuhao900914: Oh, I just realized we send the payload to api v2 eventually. Will update this to if have revenue or have price and valid quantity will pass other revenue related values.

src/main/java/com/amplitude/Event.java
yuhao900914: Sure, will add the test.

src/main/java/com/amplitude/Event.java
qingzhuozhen: `if have revenue or have price and valid quantity` this is already true when we get to this part right? What's our normal way to calculate this in SDK?

qingzhuozhen (approved):

PR #31, opened 2021-09-15
feat: add setServerUrl api

yuhao900914:
### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -99,8 +99,7 @@ public class Amplitude {
     if (updatedHttpCallMode == HttpCallMode.BATCH_HTTPCALL) {
       httpCall = new BatchHttpCall(apiKey, serverUrl != null ? serverUrl : Constants.BATCH_API_URL);
     } else {
-      httpCall =
-          new GeneralHttpCall(apiKey, serverUrl != null ? serverUrl : Constants.BATCH_API_URL);
+      httpCall = new GeneralHttpCall(apiKey, serverUrl != null ? serverUrl : Constants.API_URL);
     }
   }
 
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -17,7 +17,7 @@ public class Amplitude {
 
   private static Map<String, Amplitude> instances = new HashMap<>();
   private String apiKey;
-  private String serverUrl = Constants.API_URL;
+  private String serverUrl;
 
   private AmplitudeLog logger;
 
@@ -68,7 +68,7 @@ public class Amplitude {
    */
   public void init(String key) {
     apiKey = key;
-    updateHttpCall(HttpCallMode.REGULAR_HTTPCALL, serverUrl);
+    updateHttpCall(HttpCallMode.REGULAR_HTTPCALL);
   }
 
   /**
@@ -81,7 +81,7 @@ public class Amplitude {
     boolean isValidServerUrl = url.startsWith("http://") || url.startsWith("https://");
     if (!isValidServerUrl) return;
     serverUrl = url;
-    updateHttpCall(httpCallMode, serverUrl);
+    updateHttpCall(httpCallMode);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


haoliu-amp (changes requested):
User might pass the EU batch endpoint. It also ends with `batch` but it is `https://api.eu.amplitude.com/batch`

Refer to https://developers.amplitude.com/docs/batch-event-upload-api

With current logic, I believe it will be treated as a normal endpoint. 

cc: @qingzhuozhen

yuhao900914:
> User might pass the EU batch endpoint. It also ends with `batch` but it is `https://api.eu.amplitude.com/batch`
> 
> Refer to https://developers.amplitude.com/docs/batch-event-upload-api
> 
> With current logic, I believe it will be treated as a normal endpoint.
> 
> cc: @qingzhuozhen

@haoliu-amp 
`setServerUrl ` only used for setting amplitude endpoints, right? 
Can I use the `batch` keyword in the endpoint to determine if it's a regular request or batch request?

haoliu-amp (changes requested):
Look good to me. Could we write a fews tests for sending data to EU API and EU BATCH API?

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -81,12 +95,12 @@ public void useBatchMode(Boolean isBatchMode) {
...
-      httpCallMode = updatedHttpCallMode;
+    httpCallMode = updatedHttpCallMode;
+    if (updatedHttpCallMode == HttpCallMode.BATCH_HTTPCALL) {
+      httpCall = new BatchHttpCall(apiKey, serverUrl != null ? serverUrl : Constants.BATCH_API_URL);
+    } else {
       httpCall =
-          (updatedHttpCallMode == HttpCallMode.BATCH_HTTPCALL)
-              ? new BatchHttpCall(apiKey)
-              : new GeneralHttpCall(apiKey);
+          new GeneralHttpCall(apiKey, serverUrl != null ? serverUrl : Constants.BATCH_API_URL);
```
qingzhuozhen, line 103: The default for general is not batch url right?

src/main/java/com/amplitude/BatchHttpCall.java
```diff
@@ -1,10 +1,11 @@
 package com.amplitude;
 
 public class BatchHttpCall extends HttpCall {
-  private static String apiUrl = Constants.BATCH_API_URL;
+  private static String apiUrl;
 
-  protected BatchHttpCall(String apiKey) {
+  protected BatchHttpCall(String apiKey, String batchServerUrl) {
```
qingzhuozhen, line 6: This is ok, we can also adjust the `HttpCall` constructor and set from there.

src/main/java/com/amplitude/BatchHttpCall.java
```diff
@@ -1,10 +1,11 @@
 package com.amplitude;
 
 public class BatchHttpCall extends HttpCall {
-  private static String apiUrl = Constants.BATCH_API_URL;
+  private static String apiUrl;
```
qingzhuozhen, line 4: why we make this static earlier? I don't see we have a strong need for it.

src/main/java/com/amplitude/Amplitude.java
yuhao900914: Yes, the default is the general one.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -70,6 +71,19 @@ public void init(String key) {
     updateHttpCall(HttpCallMode.REGULAR_HTTPCALL);
   }
 
+  /**
+   * Sends events to a different URL other than Constants.API_URL or Constants.BATCH_API_URL. Used
+   * for proxy server.
+   *
+   * @param url the server URL for sending event
+   */
+  public void setServerUrl(String url) {
```
qingzhuozhen, line 80: let's add some test case for it.

src/main/java/com/amplitude/Amplitude.java
qingzhuozhen: I mean when serverUrl is null here the fallback url is batch url not the general one.

src/main/java/com/amplitude/BatchHttpCall.java
haoliu-amp: +1

haoliu-amp (changes requested):

src/main/java/com/amplitude/Amplitude.java
yuhao900914: Oh, good catch! Will fix this.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -122,6 +124,34 @@ public void testLogEventWithInvalidResponse(boolean useBatch)
...
+  public void testDefaultServerUrl(boolean useBatch)
+      throws NoSuchFieldException, IllegalAccessException {
+    Amplitude amplitude = Amplitude.getInstance("test");
+    amplitude.init(apiKey);
+    amplitude.useBatchMode(useBatch);
+
+    Field httpCallField = amplitude.getClass().getDeclaredField("httpCall");
+    httpCallField.setAccessible(true);
+    HttpCall httpCall = (HttpCall) httpCallField.get(amplitude);
+    System.out.println(httpCall.getApiUrl());
```
qingzhuozhen, line 138: clean up this?

yuhao900914: Cleaned.

qingzhuozhen (approved):

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -122,6 +124,33 @@ public void testLogEventWithInvalidResponse(boolean useBatch)
...
+    amplitude.useBatchMode(useBatch);
+
+    Field httpCallField = amplitude.getClass().getDeclaredField("httpCall");
+    httpCallField.setAccessible(true);
+    HttpCall httpCall = (HttpCall) httpCallField.get(amplitude);
+    assertEquals(httpCall.getApiUrl(), useBatch ? Constants.BATCH_API_URL : Constants.API_URL);
+  }
+
+  @Test
+  public void testSetServerUrl() throws NoSuchFieldException, IllegalAccessException {
```
haoliu-amp, line 142: Let's add a test for EU batch API

haoliu-amp (changes requested):

haoliu-amp (approved):

PR #33, opened 2021-09-29
fix: AMP-41837 session id default to -1

dantetam:
Also bumps the version number to 1.3.1. I'll release 1.3.1 here on GitHub as well as Sonatype


src/main/java/com/amplitude/Event.java
```diff
@@ -151,7 +151,7 @@
    * want to associate events with a particular system. A session_id of -1 is the same as no
    * session_id specified.
    */
-  public long sessionId;
+  public long sessionId = -1;
```
qingzhuozhen, line 154: looks good, can we add a quick test to assert the default value?

dantetam: It's been done

yuhao900914 (approved):

qingzhuozhen (approved):

PR #36, opened 2021-11-08
feat: add event callback support

qingzhuozhen:
### Summary

- add event callback support
- adjust retry on client level
- rename package in demo to another name so unit test coverage can be separate.
- add more unit tests 

### Checklist


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -96,6 +94,14 @@ public void setLogMode(AmplitudeLog.LogMode logMode) {
     this.logger.setLogMode(logMode);
   }
 
+  /**
+   * Set event callbacks which are triggered after event sent
+   * @param eventCallbacks
+   */
+  public void setEventCallbacks(List<AmplitudeEventCallback> eventCallbacks) {
```
haoliu-amp, line 101: Why would we need more than one callback?

qingzhuozhen: This was required by doordash specifically. I can double check with them. My guess is for the callbacks, they want some flexibility or can have multiple callbacks from different services.

haoliu-amp: ok. If we do accept a list of callbacks, we also need to have a func to take only one callback too.

qingzhuozhen: I see, do you mean add a method like `addEventCallback(AmplitudeEventCallback callback)` as an extra interface? And the one taken in will be added into current list? That could be an easier way then have to set the whole list each time.

qingzhuozhen: @haoliu-amp @justin-fiedler Just updated interface to `setEventCallback`

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -55,6 +57,14 @@ public void testLogEventSuccess(boolean useBatch)
               latch.countDown();
               return response;
             });
+    List<AmplitudeEventCallback> eventCallbacks = new ArrayList<>();
```
justin-fiedler, line 60: It doesn't look like this is added via setEventCallback(). Also I don't we need the array anymore.

justin-fiedler (changes requested):

src/test/java/com/amplitude/AmplitudeTest.java
qingzhuozhen: Good catch! That was accidentally left inside tests. Let me remove and clean up this test.

src/demo/java/com/demo/amplitude/DemoController.java
```diff
@@ -0,0 +1,35 @@
...
+  @RequestMapping("/")
+  public String index() {
+    Amplitude amplitude = Amplitude.getInstance("INSTANCE_NAME");
+    amplitude.init("8e07b9d451a7d07bd33f6e9ba5870f21");
+    amplitude.logEvent(new Event("Test Event", "test_user_id"));
+    amplitude.setLogMode(AmplitudeLog.LogMode.DEBUG);
+    AmplitudeEventCallback eventCallback =
+        new AmplitudeEventCallback() {
+          @Override
+          public void onEventSent(Event event, int status, String message) {
```
justin-fiedler, line 25: @qingzhuozhen I'm re-questioning the name for `onEventSent`. Not sure it is very clear what "sent" means and if it corresponds to logEvent, flush, request sent to server, or server response. Wondering if we can think of something more explicit such as `AmplitudeEventCallback.onServerResponse`

I am also thinking future customers might want additional callbacks, some that might not be per event. Maybe we can rename the interface to be more generic.

```
public interface AmplitudeCallbacks {
    void onLogEventResponse(Event event, int status, String message);
    // Example of potential future callback not related to events
    // void onFlush();
}

amplitude.addCallbacks(new AmplitudeCallbacks() { ... })
```

justin-fiedler (changes requested):

src/demo/java/com/demo/amplitude/DemoController.java
qingzhuozhen: @justin-fiedler I updated the names on class and method. Feel free to take another look. One extra thought about extending this, would each time when we extend AmplitudeCallback interface be a breaking change for upgrade?

justin-fiedler: @qingzhuozhen good point on the breaking change on each update. We could make AmplitudeCallbacks a class instead of an interface and have default noop implementations for all methods. What do you think?

qingzhuozhen: @justin-fiedler yeah, I guess switch to a class makes more sense for future change. Personally like interface more but seems that won't be best fit considering future methods. Let me switch it to class then.

qingzhuozhen: @justin-fiedler just updated interface to class.

justin-fiedler (approved):

PR #38, opened 2021-11-15
fix: thread-safe improvements

qingzhuozhen:
### Summary

- Fix issues for multi-threading case
- Set up some unit tests to help catch possible issues.
- Gathering thoughts or suggestions on the fixes or potential issues.

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -96,6 +96,7 @@ public class Amplitude {
 
   /**
    * Set event callback which are triggered after event sent
+   *
    * @param callbacks AmplitudeCallbacks or null to clean up.
    */
   public void setCallbacks(AmplitudeCallbacks callbacks) {
@@ -107,7 +108,7 @@ public class Amplitude {
    *
    * @param event The event to be sent
    */
-  public void logEvent(Event event) {
+  public synchronized void logEvent(Event event) {
     eventsToSend.add(event);
     if (eventsToSend.size() >= Constants.EVENT_BUF_COUNT) {
       flushEvents();
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -149,6 +149,7 @@ class HttpTransport {
                         numEventsRemoved += 1;
                       }
                     }
+                    triggerEventCallbacks(eventsToDrop, retryResult.statusCode, "Invalid events.");
                     eventCount -= numEventsRemoved;
                     eventsInRetry.addAndGet(-numEventsRemoved);
                   }
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -6,6 +6,7 @@ import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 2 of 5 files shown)


src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -161,11 +150,11 @@ private void retryEventsOnLoop(String userId, String deviceId) {
                   }
                   if (shouldReduceEventCount) {
                     if (!isLastTry) {
-                      eventCount /= 2;
```
justin-fiedler, line 164: Why'd you move this down?

src/test/java/com/amplitude/AmplitudeMultiThreadTest.java
```diff
@@ -0,0 +1,211 @@
...
+    CountDownLatch threadLatch = new CountDownLatch(threads);
+    ExecutorService executorService = Executors.newFixedThreadPool(threads);
+
+    for (int i = 0; i < threads; i++) {
+      executorService.execute(new LogEventRunnable(amplitude, total / threads, threadLatch));
+    }
+
+    try {
+      threadLatch.await();
+      callbacks.blockWithTimeout(15);
```
justin-fiedler, line 63: Why do you need this timeout?

src/main/java/com/amplitude/HttpTransport.java
qingzhuozhen: I realized the original way is not correct. Basically eventsBuffer is not changed in the loop or by subList. The original way we just call more event callbacks then we want to.

src/test/java/com/amplitude/AmplitudeMultiThreadTest.java
qingzhuozhen: We need this timeout so when there is any issues the tests won't just hang there..

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -107,7 +108,7 @@ public void setCallbacks(AmplitudeCallbacks callbacks) {
    *
    * @param event The event to be sent
    */
-  public void logEvent(Event event) {
+  public synchronized void logEvent(Event event) {
```
qingzhuozhen, line 111: The goal for this is between add and clear we don't lose events..

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -308,9 +292,49 @@ private void triggerEventCallbacks(List<Event> events, int status, String messag
     if (callbacks == null || events == null || events.isEmpty()) {
       return;
     }
-
     for (Event event : events) {
       callbacks.onLogEventServerResponse(event, status, message);
     }
   }
+
+  private void addEventToBuffer(String userId, String deviceId, Event event) {
```
qingzhuozhen, line 300: The goal here is for add or remove we don't lose events on the buffer.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -231,7 +220,7 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
           || events.size() == 1) {
         // Return early if there's an issue with the entire payload
         // or if there's only one event and its invalid
-        triggerEventCallbacks(eventsToDrop, response.code, response.error);
+        triggerEventCallbacks(events, response.code, response.error);
```
qingzhuozhen, line 223: Covers some callbacks weren't triggered

src/test/java/com/amplitude/AmplitudeMultiThreadTest.java
justin-fiedler: Oh got it. I thought it was an additional wait, not a timeout. Makes sense.

justin-fiedler (approved):

PR #43, opened 2022-01-09
Add options key to sent message.

hanangi:
feat: add options key to sent event.

### Summary

Add option key to sent data.. In particular, to be used to override min_id_length option...

### Checklist

* Does your PR have a breaking change?: no

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -1,5 +1,7 @@
 package com.amplitude;
 
+import org.json.JSONObject;
+
 import java.util.*;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
@@ -16,6 +18,11 @@ public class Amplitude {
   private HttpCallMode httpCallMode;
   private HttpCall httpCall;
   private HttpTransport httpTransport;
+
+  /**
+   * A dictionary of key-value pairs that represent additional instructions for server save operation.
+   */
+  private JSONObject options;
 
   /**
    * Private internal constructor for Amplitude. Please use `getInstance(String name)` or
@@ -72,6 +79,13 @@ public class Amplitude {
     boolean isValidServerUrl = url.startsWith("http://") || url.startsWith("https://");
     if (!isValidServerUrl) return;
     serverUrl = url;
+    updateHttpCall(httpCallMode);
+  }
+
+  public void addOption(String key, Object value) {
+    if(options == null) options = new JSONObject();
+
+    options.put(key, value);
     updateHttpCall(httpCallMode);
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -17,6 +19,11 @@
   private HttpCall httpCall;
   private HttpTransport httpTransport;
 
+  /**
+   * A dictionary of key-value pairs that represent additional instructions for server save operation.
+   */
+  private JSONObject options;
```
justin-fiedler, line 25:
```suggestion
  private Options options;
```

Would prefer this to be strongly typed similar to [Event](https://github.com/amplitude/Amplitude-Java/blob/main/src/main/java/com/amplitude/Event.java).

```
// Options.java
public class Options {
  /** A unique identifier for your event. */
  public String minIdLength;

 public JSONObject toJsonObject() { ... } => { "min_id_length": minIdLength };
}
```

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -75,6 +82,11 @@ public void setServerUrl(String url) {
     updateHttpCall(httpCallMode);
   }
 
+  public void setOptions(JSONObject options) {
```
justin-fiedler, line 85:
```suggestion
  public void setOptions(Options options) {
```

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -41,6 +43,7 @@ protected Response makeRequest(List<Event> events) throws AmplitudeInvalidAPIKey
 
       JSONObject bodyJson = new JSONObject();
       bodyJson.put("api_key", this.apiKey);
+      if(options != null) bodyJson.put("options", options);
```
justin-fiedler, line 46:
```suggestion
      if(options != null) bodyJson.put("options", options.toJsonObject());
```

justin-fiedler (changes requested):

justin-fiedler:
Hi @hanangi really appreciate the PR!

hanangi:
HI Justin... hopefully you'll like it.
Hanan.

src/main/java/com/amplitude/Options.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class Options {
+
+    /**
+     * Minimum length for user ID or device ID value.
+     */
+    public String minIdLength;
```
qingzhuozhen, line 11: Would it better to make this an Integer?

hanangi: Sure! fixing.

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -14,10 +13,12 @@
 public class HttpCall {
   private final String apiKey;
   private final String serverUrl;
+  private final Options options;
 
-  protected HttpCall(String apiKey, String serverUrl) {
+  protected HttpCall(String apiKey, String serverUrl, Options options) {
```
justin-fiedler, line 18: Since Options are optional can you please add a convenience constructor. This will also make it backwards compatible with previous versions.

```
  protected HttpCall(String apiKey, String serverUrl) {
    this(apiKey, serverUrl, null)
  }
```

justin-fiedler (changes requested):
@hanangi this looks really good. Thanks for the updates. I think it will be good to merge after adding the convenience constructor to HttpCall.

src/main/java/com/amplitude/HttpCall.java
hanangi: Done.

justin-fiedler (approved):
Thanks @hanangi! Looks great.

qingzhuozhen (approved):

PR #44, opened 2022-01-11
fix: Log every possible status code

bohan-amplitude:
### Summary

In HttpTransport class, sendEventWithRetry method, adding failed, timeout and other unknown status callback logs.

### Checklist

* [X] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Java/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?

### DIFF
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -210,6 +210,9 @@ class HttpTransport {
       case PAYLOAD_TOO_LARGE:
         shouldRetry = true;
         break;
+      case TIMEOUT:
+          triggerEventCallbacks(events, response.code, response.error);
+        break;
       case INVALID:
         if (events.size() == 1) {
           shouldRetry = false;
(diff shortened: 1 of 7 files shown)


src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -206,6 +210,9 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
       case PAYLOAD_TOO_LARGE:
         shouldRetry = true;
         break;
+      case TIMEOUT:
+          triggerEventCallbacks(events, response.code, response.error);
```
qingzhuozhen, line 214: Why are we trigger callbacks here? I don't see shouldRetry is updated, so this is still going to retry again right?

qingzhuozhen: I wonder in this case we would get more callbacks than expected. The error callbacks are more calling once after either success or we don't retry anymore

qingzhuozhen (changes requested):

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: removed the two line and changed the test for this

PR #45, opened 2022-01-17
feat: support event with callback

qingzhuozhen:
### Summary
add backs on event level

### Checklist

* Does your PR have a breaking change?: no

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -4,13 +4,14 @@ on: [pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     steps:
       - name: Set up JDK 8
         uses: actions/setup-java@v2
         with:
           java-version: '8'
           distribution: 'adopt'
+          architecture: x64
       - uses: actions/checkout@v2
       - name: Test with Gradle
         run: gradle test:test
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -10,7 +10,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: '8'
-          distribution: 'adopt'
+          distribution: 'temurin'
       - uses: actions/checkout@v2
       - name: Test with Gradle
         run: gradle test:test --stacktrace
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -13,5 +13,5 @@ jobs:
           distribution: 'adopt'
       - uses: actions/checkout@v2
       - name: Test with Gradle
-        run: gradle test:test
+        run: gradle test:test --stacktrace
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.*;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Future;
```
yuhao900914, line 5: Why do we need to import this?

qingzhuozhen: Good catch, let me remove it. This was added when I play around the suggestion to return a Future and later we picked another way.

.github/workflows/pull-request-test.yml
```diff
@@ -4,13 +4,14 @@ on: [pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
justin-fiedler, line 7: Do we need macos? [Costs 10x](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates) compared to Ubuntu.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -311,11 +311,14 @@ protected boolean shouldRetryForStatus(Status status) {
   }
 
   private void triggerEventCallbacks(List<Event> events, int status, String message) {
-    if (callbacks == null || events == null || events.isEmpty()) {
+    if (events == null || events.isEmpty()) {
       return;
     }
     for (Event event : events) {
-      callbacks.onLogEventServerResponse(event, status, message);
+      AmplitudeCallbacks eventCallback = event.callback != null ? event.callback : callbacks;
```
justin-fiedler, line 318: I think should we still call the client level callbacks here in addition to the per-event callback

justin-fiedler (changes requested):

.github/workflows/pull-request-test.yml
qingzhuozhen: That is copied from @bohan-amplitude 's change to make the tests more stable.. @bohan-amplitude maybe see if we can make this more stable with ubuntu?

src/main/java/com/amplitude/HttpTransport.java
qingzhuozhen: We can do that too, I was trying to debate if we need to do replace or in addition. Would in addition sounds more natural for the intention?

.github/workflows/pull-request-test.yml
qingzhuozhen: let revert this for now.

src/main/java/com/amplitude/HttpTransport.java
qingzhuozhen: Updated to call extra now.

bohan-amplitude (approved):

PR #49, opened 2022-01-25
feat: add print debug info to console for client flush and event sending with response detail

bohan-amplitude:
### Summary

When logger set to debug mode:
print to console how many events are flushed by Amplitude client
print to console response detail for events sending requests
print to console response detail for events retry requests

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -174,6 +174,7 @@ public class Amplitude {
     if (eventsToSend.size() > 0) {
       List<Event> eventsInTransit = new ArrayList<>(eventsToSend);
       eventsToSend.clear();
+      logger.log(logger.getTimeStamp(), eventsInTransit.size() + " events to flush");
       httpTransport.sendEventsWithRetry(eventsInTransit);
     }
   }
@@ -205,5 +206,17 @@ public class Amplitude {
               });
       flushThread.start();
     }
+  }
+
+  public void debug(String message) {
+    logger.log(logger.getTimeStamp(), message);
+  }
+
+  public void warn(String message) {
+    logger.warn(logger.getTimeStamp(), message);
+  }
+
+  public void error(String message) {
+    logger.error(logger.getTimeStamp(), message);
   }
 }
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -174,7 +174,6 @@ public class Amplitude {
     if (eventsToSend.size() > 0) {
       List<Event> eventsInTransit = new ArrayList<>(eventsToSend);
       eventsToSend.clear();
-      logger.log(logger.getTimeStamp(), eventsInTransit.size() + " events to flush");
       httpTransport.sendEventsWithRetry(eventsInTransit);
     }
   }
@@ -206,17 +205,5 @@ public class Amplitude {
               });
       flushThread.start();
     }
-  }
-
-  public void debug(String message) {
-    logger.log(logger.getTimeStamp(), message);
-  }
-
-  public void warn(String message) {
-    logger.warn(logger.getTimeStamp(), message);
-  }
-
-  public void error(String message) {
-    logger.error(logger.getTimeStamp(), message);
   }
 }
--- a/src/main/java/com/amplitude/AmplitudeLog.java
+++ b/src/main/java/com/amplitude/AmplitudeLog.java
@@ -8,10 +8,6 @@ public class AmplitudeLog {
   private LogMode logMode = LogMode.ERROR;
   private static SimpleDateFormat sdfDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
 
-  public static String getTimeStamp(){
-    return sdfDate.format(new Date());
-  }
-
   public void setLogMode(LogMode logMode) {
     this.logMode = logMode;
   }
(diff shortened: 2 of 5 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -206,4 +207,16 @@ private void scheduleFlushEvents() {
       flushThread.start();
     }
   }
+
+  public void debug(String message) {
+    logger.log(logger.getTimeStamp(), message);
```
haoliu-amp, line 212: Why not just `log(message)`? What is the point of exposing the time as a parameter?

src/main/java/com/amplitude/Response.java
```diff
@@ -107,4 +107,19 @@ protected boolean isUserOrDeviceExceedQuote(String userId, String deviceId) {
     Collections.sort(invalidIndices);
     return invalidIndices;
   }
+
+  public String toString() {
+    JSONObject json = new JSONObject();
+    json.put("code", this.code);
+    json.put("status", this.status.name());
+    if (this.error != null)
```
haoliu-amp, line 115: Would recommend to have curly bracket around the logic for if/else

haoliu-amp (changes requested):

src/main/java/com/amplitude/Amplitude.java
bohan-amplitude: I thought timestamp can help user further analyze what happened during certain period of time, like average logged/flushed events/second. I found that the logger.error was used like logger.error("Invalid API Key", exception.getMessage()) . Should I keep the same format here without timestamp?  Two parameters for log method may not be necessary.

haoliu-amp: When you log a message, I would assume the timestamp is always just the current time, right? I would say passing a customized message is more useful.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -191,6 +197,12 @@ private void retryEventsOnLoop(String userId, String deviceId) {
   private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<Event> events)
       throws AmplitudeInvalidAPIKeyException {
     Response response = httpCall.makeRequest(events);
+      if (logger.getLogMode() == AmplitudeLog.LogMode.DEBUG) {
+          StringBuilder debugMessage = new StringBuilder();
+          debugMessage.append("Retry for userId ").append(userId).append(", deviceId ").append(deviceId)
```
haoliu-amp, line 202: Could we just use string format instead here? That will look cleaner.

haoliu-amp (changes requested):

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: Sure, since concatenating all in one place, there's no performance difference.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -174,6 +174,7 @@ public synchronized void flushEvents() {
     if (eventsToSend.size() > 0) {
       List<Event> eventsInTransit = new ArrayList<>(eventsToSend);
       eventsToSend.clear();
+      logger.log(logger.getTimeStamp(), eventsInTransit.size() + " events to flush");
```
qingzhuozhen, line 177: I hope the earlier method was called debug other than log, two log method is a bit confusing.

src/main/java/com/amplitude/Amplitude.java
```diff
@@ -206,4 +207,16 @@ private void scheduleFlushEvents() {
       flushThread.start();
     }
   }
+
+  public void debug(String message) {
+    logger.log(logger.getTimeStamp(), message);
+  }
+
+  public void warn(String message) {
```
qingzhuozhen, line 215: What's the use case for these new methods? Do we really need them?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -89,6 +89,10 @@ public void setCallbacks(AmplitudeCallbacks callbacks) {
           Response response = null;
           try {
             response = httpCall.makeRequest(events);
+            if (logger.getLogMode() == AmplitudeLog.LogMode.DEBUG) {
```
qingzhuozhen, line 92: We already have a logMode check right? Then we don't need to do this check here. I was thinking you can just call logger.debug(...)

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -89,6 +89,10 @@ public void setCallbacks(AmplitudeCallbacks callbacks) {
...
+                String debugMessage = "Sending " + events.size() + " events. \n" + response.toString();
+                logger.log(logger.getTimeStamp(), debugMessage);
```
qingzhuozhen, line 94: The tag usage maybe from Android, show class or more meaningful information. I guess we don't need to use timestamp as a tag. If you prefer to have timestamp in debug message, how about just change format inside the logger debug method?

src/main/java/com/amplitude/AmplitudeLog.java
```diff
@@ -1,12 +1,24 @@
...
+
+  public static String getTimeStamp(){
+    return sdfDate.format(new Date());
+  }
 
   public void setLogMode(LogMode logMode) {
     this.logMode = logMode;
   }
 
+  public LogMode getLogMode() {
```
qingzhuozhen, line 18: I believe we don't need this once we make other changes.

qingzhuozhen (changes requested):

src/main/java/com/amplitude/Amplitude.java
bohan-amplitude: Agree, also I think what Qingzhuo mentioned make sense, we don't need to expose these log interface to customers.

qingzhuozhen (approved):
lgtm

PR #52, opened 2022-01-28
feat: add a demo for customer who wants to send large amount of historical events using batch mode

bohan-amplitude:
### Summary

1.  add an example in Demo package src/demo/java/com/demo/amplitude/LocalUploadDemo.java showing a use case upload bulk amount of events locally with a script using java SDK.
2. Amplitude client: add a method shouldWait(event), return true if the client was busy or event's userId or deviceId was throttled recently.
3. Amplitude client: change logEvent and flushEvents to non synchronized. Manage events queue multithread access by object lock.
4. Change scheduleFlushEvents method to synchronized to prevent creation of multiple flush threads.
5. HttpTransport class: retry using a thread pool.
6. HttpTransport class: fix issue of missing callbacks for events dropped because of retry buffer size limit.
7. HttpTransport class: fix error of counting events in buffer
8. HttpTransport class: fix bug of not reduce payload size of payload too large response
9. HttpTransport class: fix bug of dropping wrong events in the retry process.

### Checklist

* Does your PR have a breaking change?: yes

### DIFF
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -373,4 +373,8 @@ class HttpTransport {
         });
     return eventQueues.isEmpty() ? null : eventQueues.get(0);
   }
+
+  public boolean shoudWait() {
+      return eventsInRetry.intValue() >= Constants.MAX_CACHED_EVENTS;
+  }
 }
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -137,6 +137,7 @@ class HttpTransport {
               if (eventsBuffer == null || eventsBuffer.size() == 0) {
                 return;
               }
+              eventsInRetry.addAndGet(-eventsBuffer.size());
               int retryTimes = Constants.RETRY_TIMEOUTS.length;
               for (int numRetries = 0; numRetries < retryTimes; numRetries++) {
                 int eventCount = eventsBuffer.size();
@@ -151,10 +152,8 @@ class HttpTransport {
                   boolean shouldRetry = retryResult.shouldRetry;
                   if (!shouldRetry) {
                     // call back done in retryEventsOnce
-                    eventsInRetry.addAndGet(-eventCount);
                     break;
                   } else if (isLastTry) {
-                    eventsInRetry.addAndGet(-eventCount);
                     triggerEventCallbacks(eventsBuffer, retryResult.statusCode, "Event retries exhausted.");
                     break;
                   }
@@ -171,10 +170,8 @@ class HttpTransport {
                       }
                     }
                     triggerEventCallbacks(eventsToDrop, retryResult.statusCode, "Invalid events.");
-                    eventsInRetry.addAndGet(-numEventsRemoved);
                   } else if (shouldReduceEventCount) {
                     List<Event> eventsToDrop = eventsBuffer.subList(eventCount / 2, eventCount);
-                    eventsInRetry.addAndGet(-eventsToDrop.size());
                     triggerEventCallbacks(eventsToDrop, retryResult.statusCode, "Event dropped for retry");
                     eventsBuffer = eventsBuffer.subList(0, eventCount / 2);
                   }
@@ -183,6 +180,7 @@ class HttpTransport {
                   // The retry logic should only be executed after the API key checking passed.
                   // This catch AmplitudeInvalidAPIKeyException is just for handling
                   // retryEventsOnce in thread.
+                  logger.debug("RETRY", "Retry thread got interrupted");
                   Thread.currentThread().interrupt();
                 }
               }
@@ -282,7 +280,7 @@ class HttpTransport {
                 throttledDeviceId.put(event.deviceId, throttledDevice.getInt(event.deviceId));
               }
             } catch (JSONException e) {
-              logger.debug("ThROTTLED", "Error get throttled userId or deviceId");
+              logger.debug("THROTTLED", "Error get throttled userId or deviceId");
             }
           }
         } else {
@@ -324,7 +322,7 @@ class HttpTransport {
     }
   }
 
-  private void addEventToBuffer(String userId, String deviceId, Event event) {
+  private synchronized void addEventToBuffer(String userId, String deviceId, Event event) {
     idToBuffer.compute(
         userId,
         (key, value) -> {
@@ -338,6 +336,7 @@ class HttpTransport {
                   deviceValue = new ConcurrentLinkedQueue<>();
                 }
                 deviceValue.add(event);
+                eventsInRetry.incrementAndGet();
                 return deviceValue;
               });
           return value;
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -344,7 +344,7 @@ class HttpTransport {
         });
   }
 
-  private List<Event> getEventsFromBuffer(String userId, String deviceId) {
+  private synchronized List<Event> getEventsFromBuffer(String userId, String deviceId) {
     if (idToBuffer.containsKey(userId) && idToBuffer.get(userId).containsKey(deviceId)){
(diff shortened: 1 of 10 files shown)


justin-fiedler (changes requested):

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -207,10 +216,11 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
           triggerEventCallbacks(events, response.code, response.error);
         }
         // Reduce the payload to reduce risk of throttling
-        shouldReduceEventCount = true;
+        //shouldReduceEventCount = true;
```
qingzhuozhen, line 219: why comment out?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -228,86 +246,57 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
...
+        triggerEventCallbacks(events, response.code, "Unknown response status.");
+        return eventsToRetry;
     } else if (response.status == Status.INVALID) {
-      if ((response.invalidRequestBody != null
-              && response.invalidRequestBody.has("missingField")
-              && response.invalidRequestBody.getString("missingField").length() > 0)
-          || events.size() == 1) {
-        // Return early if there's an issue with the entire payload
-        // or if there's only one event and its invalid
-        triggerEventCallbacks(events, response.code, response.error);
```
qingzhuozhen, line 243: Support for this case is removed?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -32,17 +33,27 @@ protected EventsRetryResult(
...
+  private Object throttleLock = new Object();
+  private Map<String, Integer> throttledUserId = new HashMap<>();
+  private Map<String, Integer> throttledDeviceId = new HashMap<>();
+  private ExecutorService consExec = Executors.newFixedThreadPool(1);
+  private boolean recordThrottledId = false;
 
   private HttpCall httpCall;
   private AmplitudeLog logger;
   private AmplitudeCallbacks callbacks;
+  private int callb = 0;
```
qingzhuozhen, line 49: these  are used for?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -228,86 +246,57 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
   }
 
   private List<Event> getEventListToRetry(List<Event> events, Response response) {
-    List<Event> eventsToRetry = events;
+    List<Event> eventsToRetry = new ArrayList<>();
     List<Event> eventsToDrop = new ArrayList<>();
     // Filter invalid event out based on the response code.
     if (response.status == Status.SUCCESS) {
-      return new ArrayList<>();
+        triggerEventCallbacks(events, response.code, "Events sent success.");
```
qingzhuozhen, line 253: I don't think we will hit on this case actually.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -155,36 +170,30 @@ private void retryEventsOnLoop(String userId, String deviceId) {
...
                   // This catch AmplitudeInvalidAPIKeyException is just for handling
                   // retryEventsOnce in thread.
+                  logger.debug("RETRY", "Retry thread got interrupted");
                   Thread.currentThread().interrupt();
                 }
               }
-              eventsInRetry.addAndGet(-eventCount);
+              if (recordThrottledId) {
+                synchronized (throttleLock) {
+                  if (throttledUserId.containsKey(userId)) {
```
qingzhuozhen, line 189: The throttling logic here is in
getEventListToRetry(initial failure) -> add userId, deviceId -> wait 100ms -> start retry -> remove userId, deviceId from throttle after retry success or exhaust retries
Is that right?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -334,43 +323,71 @@ private void triggerEventCallbacks(List<Event> events, int status, String messag
...
+
+  public void setRecordThrottledId(boolean record) {
+    recordThrottledId = record;
+  }
+
+  class EventBufferComsumer implements Runnable {
+
+    @Override
+    public void run() {
+      while (true) {
```
qingzhuozhen, line 368: Would this thread block program exit? Or we need guidance for that?

src/test/java/com/amplitude/AmplitudeMultiThreadTest.java
```diff
@@ -2,8 +2,8 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.anyList;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
```
qingzhuozhen, line 6: is this used anywhere?

qingzhuozhen (commented):
Much changes included, let's test carefully and think this through.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: something for debug, will remove

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: I found that shouldReduceEventCount was set to true for rate limit status but not payload too large status. Is this the designed behavior same on other sdks? For rate limit(throttled) situation, as I read from http api documents, events could be retry after a short period of time. Do we just drop half of the events list here without retry?

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: I think you are right. I left it unchanged because it's there before and doesn't make things wrong. I'll remove it.

src/main/java/com/amplitude/HttpTransport.java
yuhao900914: That's the expected behavior. We have the same logic in Amplitude-Node. When it reaches the rate limit we reduce the event count for further retry. 
https://github.com/amplitude/Amplitude-Node/blob/7bb74e8d1b070cd0e5a940fac8661dde4c060e6a/packages/node/src/retry/defaultRetry.ts#L208

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: Still support this case. It's covered in the current code.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: I'll add a boolean switch to control this thread and add a method in client class to shutdown client ( flush all events, wait retry to finish, wait the consumer thread to join)

src/test/java/com/amplitude/AmplitudeMultiThreadTest.java
bohan-amplitude: Added when debug. Removed

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: I think our 'cut the half of list to drop' approach is too rough. We can retry first half and second half separately for payload too large situation or even check payload size before send out to http api.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: Changed back to trigger retry when response of first events sending attempt has certain status.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: Added back the special case

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: Added back this setting for alignment. Need to discuss the retry logic for new SDKs.

PR #53, opened 2022-02-02
feat: add a demo for customer who wants to send large amount of historical events using batch mode

bohan-amplitude:
### Summary

1. add an example in Demo package src/demo/java/com/demo/amplitude/LocalUploadDemo.java showing a use case upload bulk amount of events locally with a script using java SDK.
2. Amplitude client: add a method shouldWait(event), return true if the client was busy or event's userId or deviceId was throttled recently.
3. Amplitude client: add a method setRecordThrottledId(boolean). Config whether shouldWait() method use recent throttled userId/deviceId or not.

### Checklist

### DIFF
--- a/src/demo/java/com/demo/amplitude/LocalUploadDemo.java
+++ b/src/demo/java/com/demo/amplitude/LocalUploadDemo.java
@@ -10,46 +10,50 @@ import java.util.concurrent.TimeUnit;
 
 public class LocalUploadDemo {
 
-    public static void main(String[] args) throws InterruptedException{
-        // Create and initialize Amplitude client
-        Amplitude client = Amplitude.getInstance();
-        client.init("");
+  public static void main(String[] args) throws InterruptedException {
+    // Create and initialize Amplitude client
+    Amplitude client = Amplitude.getInstance();
+    client.init("");
 
-        // use batch mode for higher throttling limits
-        client.useBatchMode(true);
+    // use batch mode for higher throttling limits
+    client.useBatchMode(true);
 
-        // this config can print debug info into console, including response body of each requests
-        client.setLogMode(AmplitudeLog.LogMode.DEBUG);
+    // this config can print debug info into console, including response body of each requests
+    client.setLogMode(AmplitudeLog.LogMode.DEBUG);
 
-        // for large amount of events to send, config a higher update threshold to upload more events in one request
-        // payload size limit for batch api is 20MB, max events per request is 2000
-        // https://developers.amplitude.com/docs/batch-event-upload-api#feature-comparison-between-httpapi-2httpapi--batch
-        client.setEventUploadThreshold(1500);
+    // for large amount of events to send, config a higher update threshold to upload more events in
+    // one request
+    // payload size limit for batch api is 20MB, max events per request is 2000
+    // https://developers.amplitude.com/docs/batch-event-upload-api#feature-comparison-between-httpapi-2httpapi--batch
+    client.setEventUploadThreshold(1500);
 
-        // config client to record throttled userId and deviceId. shouldWait(event) will return true if userid or deviceid was throttled
-        // can wait a short period of time and continue
-        client.setRecordThrottledId(true);
-        AmplitudeCallbacks callback = new AmplitudeCallbacks() {
-            @Override
-            public void onLogEventServerResponse(Event event, int status, String message) {
-                // call back functions here
-                System.out.println(event.eventType + " " + event.userId + " " + status + " " + message);
-            }
+    // config client to record throttled userId and deviceId. shouldWait(event) will return true if
+    // userid or deviceid was throttled
+    // can wait a short period of time and continue
+    client.setRecordThrottledId(true);
+    AmplitudeCallbacks callback =
+        new AmplitudeCallbacks() {
+          @Override
+          public void onLogEventServerResponse(Event event, int status, String message) {
+            // call back functions here
+            System.out.println(event.eventType + " " + event.userId + " " + status + " " + message);
+          }
         };
-        client.setCallbacks(callback);
-        for (int i = 0; i < 10000000; i++){
-            Event ampEvent = new Event("General"+ (i % 20), "Test_UserID_B" + (i % 5000));
-            while (client.shouldWait(ampEvent)) {
-                System.out.println("Client is busy. Waiting for log event " + ampEvent.eventType);
-                TimeUnit.SECONDS.sleep(60L);
-            }
-            ampEvent.userProperties = new JSONObject()
-                    .put("property1", "p"+i)
-                    .put("property2", "p"+i)
-                    .put("property3", "p"+i)
-                    .put("property4", "p"+i)
-                    .put("property5", "p"+i);
-            client.logEvent(ampEvent);
-        }
+    client.setCallbacks(callback);
+    for (int i = 0; i < 10000000; i++) {
+      Event ampEvent = new Event("General" + (i % 20), "Test_UserID_B" + (i % 5000));
+      while (client.shouldWait(ampEvent)) {
+        System.out.println("Client is busy. Waiting for log event " + ampEvent.eventType);
+        TimeUnit.SECONDS.sleep(60L);
+      }
+      ampEvent.userProperties =
+          new JSONObject()
+              .put("property1", "p" + i)
+              .put("property2", "p" + i)
+              .put("property3", "p" + i)
+              .put("property4", "p" + i)
+              .put("property5", "p" + i);
+      client.logEvent(ampEvent);
     }
+  }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -215,6 +215,29 @@ public synchronized void flushEvents() {
...
+   * 1. Events sit in memory waiting for flush are more than flush threshold
+   * 2. Events in retry buffer are more than 16,000
+   * 3. When setRecordThrottledId(true), the userId or deviceId of the input event was throttled and retry attempt for the userId/deviceId not finished.
+   * Can be called before logEvent(event) and add some wait time if return true.
+   *
+   * @param event The event to be sent.
+   * @return true if client is busy or event may be throttled.
+   */
+  public boolean shouldWait(Event event) {
+    return eventsToSend.size() > eventUploadThreshold || httpTransport.shouldWait(event);
```
qingzhuozhen, line 229: can we add some tests for the new method?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -263,6 +279,22 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
           eventsToRetry.add(event);
         } else {
           eventsToDrop.add(event);
+          if (recordThrottledId) {
+            try {
+              JSONObject throttledUser = response.rateLimitBody.getJSONObject("exceededDailyQuotaUsers");
```
qingzhuozhen, line 284: The current throttling information is based on user id and device id. Is there any other cases for it? like reach day limit? How would we handle for that case?

src/main/java/com/amplitude/Amplitude.java
qingzhuozhen: Will the first part of logic happen? We can keep it just wonder if that would happen.

qingzhuozhen (changes requested):

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: 429 error like event per second limit or daily limit all related to  and limited by userId and deviceId. The detail error message may be different but userId/deviceId is the way to tell whether an event will be throttled.

src/main/java/com/amplitude/Amplitude.java
bohan-amplitude: Test added. The first part may happen for future multithreading handling change.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +339,27 @@ public void testWithSwallowMiddleware()
     assertEquals("middleware_user_2", sentEvent.userId);
   }
 
+  @Test
+  public void testShouldWait() throws NoSuchFieldException, IllegalAccessException {
```
qingzhuozhen, line 343: This is for unit level in client. Can we have a test to use the mocked response? Return a mocked response with throttling user and assert shouldWait true then to false?

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +339,27 @@ public void testWithSwallowMiddleware()
...
+    Amplitude amplitude = Amplitude.getInstance();
+    amplitude.init(apiKey);
+    Event event = new Event("test event", "test-user-0");
+    Event event2 = new Event("test event", "test-user-1");
+    assertFalse(amplitude.shouldWait(event));
+    Field httpTransportField = amplitude.getClass().getDeclaredField("httpTransport");
+    httpTransportField.setAccessible(true);
+    HttpTransport transport = (HttpTransport) httpTransportField.get(amplitude);
+    amplitude.setRecordThrottledId(true);
```
qingzhuozhen, line 352: Do we need to set this setting first?

src/demo/java/com/demo/amplitude/LocalUploadDemo.java
```diff
@@ -0,0 +1,55 @@
...
+import com.amplitude.Event;
+import org.json.JSONObject;
+
+import java.util.concurrent.TimeUnit;
+
+public class LocalUploadDemo {
+
+    public static void main(String[] args) throws InterruptedException{
+        // Create and initialize Amplitude client
+        Amplitude client = Amplitude.getInstance();
```
qingzhuozhen, line 15: nit: could use formatter on this file

src/test/java/com/amplitude/AmplitudeTest.java
bohan-amplitude: Yes amplitude.setRecordThrottledId(true) is a config setting to record ids with 429 response and use the ids for shouldWait() method

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: put the code block in the right place, also correct the json keywords from exceededDailyQuotaUsers to throttledUsers,  exceededDailyQuotaDevices to throttledDevices

src/test/java/com/amplitude/AmplitudeTest.java
bohan-amplitude: The first assertFalse I put their on purpose. It should be false before set the setRecordThrottledId config. I added another assertFalse after config.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +339,49 @@ public void testWithSwallowMiddleware()
...
+            invocation -> {
+              return response;
+            });
+    assertFalse(amplitude.shouldWait(event));
+    Field httpTransportField = amplitude.getClass().getDeclaredField("httpTransport");
+    httpTransportField.setAccessible(true);
+    HttpTransport transport = (HttpTransport) httpTransportField.get(amplitude);
+    amplitude.setRecordThrottledId(true);
+    assertFalse(amplitude.shouldWait(event));
+    Field userIdMapField = transport.getClass().getDeclaredField("throttledUserId");
```
qingzhuozhen, line 372: If we use the response then there is no need to use reflection right?

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +339,49 @@ public void testWithSwallowMiddleware()
...
+    Map<String, Integer> userMap = mock(HashMap.class);
+    userIdMapField.setAccessible(true);
+    userIdMapField.set(transport, userMap);
+    when(userMap.containsKey("test-user-0")).thenReturn(true);
+    when(userMap.containsKey("test-user-1")).thenReturn(false);
+    assertTrue(amplitude.shouldWait(event));
+    assertFalse(amplitude.shouldWait(event2));
+    transport.retryEvents(Arrays.asList(event, event2), response);
+    Thread.sleep(1000L);
```
qingzhuozhen, line 381: Try not use Thread sleep in test.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +339,49 @@ public void testWithSwallowMiddleware()
...
+    verify(userMap, atLeast(1)).put(eq("test-user-0"), anyInt());
```
qingzhuozhen, line 382: Assert put is not that ideal though..

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +341,51 @@ public void testWithSwallowMiddleware()
...
+              barrier.await();
+              return response;
+            });
+    assertFalse(amplitude.shouldWait(event));
+    Field httpTransportField = amplitude.getClass().getDeclaredField("httpTransport");
+    httpTransportField.setAccessible(true);
+    HttpTransport transport = (HttpTransport) httpTransportField.get(amplitude);
+    amplitude.setRecordThrottledId(true);
+    assertFalse(amplitude.shouldWait(event));
+    transport.retryEvents(Arrays.asList(event, event2), response);
```
qingzhuozhen, line 382: Looks much better now! nit: This can be initiated from amplitude actually, just add another response for it.

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -335,6 +341,51 @@ public void testWithSwallowMiddleware()
...
+    assertTrue(latch.await(1, TimeUnit.SECONDS));
+    assertTrue(amplitude.shouldWait(event));
+    assertFalse(amplitude.shouldWait(event2));
+    barrier.await();
```
qingzhuozhen, line 386: Nit: anything we can think of after do `barrier.await()`? Since we release the barrier now, maybe in callback we can assert the shouldWait to be false.

qingzhuozhen (approved):
Lgtm now!

PR #55, opened 2022-02-02
fix: issue of possible drop wrong event and error counting events in retry buffer

bohan-amplitude:
### Summary

1. add callback for events dropped because of events buffer size limit
2. remove prune method. This method remove events from list and the list will later be used by looking for events by old list index. Result in drop wrong events of index out of bound exception
3. Increase counter in addEventToBuffer method

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -230,6 +230,7 @@ class HttpTransport {
         break;
       case PAYLOAD_TOO_LARGE:
         shouldRetry = true;
+        shouldReduceEventCount = true;
         break;
       case INVALID:
         if (events.size() == 1) {
@@ -238,6 +239,14 @@ class HttpTransport {
         } else {
           eventIndicesToRemove = response.collectInvalidEventIndices();
         }
+        break;
+      case UNKNOWN:
+        shouldRetry = false;
+        triggerEventCallbacks(events, response.code, "Unknown response status.");
+        break;
+      case FAILED:
+        shouldRetry = false;
+        triggerEventCallbacks(events, response.code, "Event sent Failed.");
         break;
       default:
         break;
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -75,10 +75,13 @@ class HttpTransport {
 
   // The main entrance for the retry logic.
   public void retryEvents(List<Event> events, Response response) {
-    List<Event> eventsToSend = pruneEvent(events);
-    if (eventsInRetry.intValue() < Constants.MAX_CACHED_EVENTS) {
-      onEventsError(eventsToSend, response);
-    }
+      if (eventsInRetry.get() < Constants.MAX_CACHED_EVENTS) {
+          onEventsError(events, response);
+      } else {
+          String message = "Retry buffer is full, events dropped.";
+          logger.warn("DROP EVENTS", message);
+          triggerEventCallbacks(events, response.code, message);
+      }
   }
 
   public void setHttpCall(HttpCall httpCall) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -75,10 +75,13 @@ public void sendEventsWithRetry(List<Event> events) {
...
   // The main entrance for the retry logic.
   public void retryEvents(List<Event> events, Response response) {
-    List<Event> eventsToSend = pruneEvent(events);
-    if (eventsInRetry.intValue() < Constants.MAX_CACHED_EVENTS) {
-      onEventsError(eventsToSend, response);
-    }
+      if (eventsInRetry.get() < Constants.MAX_CACHED_EVENTS) {
+          onEventsError(events, response);
+      } else {
+          String message = "Retry buffer is full, events dropped.";
```
qingzhuozhen, line 81: Include the events count in the message? The log info doesn't tell much now.

src/test/java/com/amplitude/HttpTransportTest.java
```diff
@@ -95,8 +95,10 @@ public void onLogEventServerResponse(Event event, int status, String message) {
     assertTrue(latch.await(1L, TimeUnit.SECONDS));
     verify(httpCall, times(4)).makeRequest(anyList());
     for (int i = 0; i < events.size(); i++) {
-      if (i < (events.size() / 2)) {
+      if (i < (events.size() / 4)) {
         assertEquals(200, resultMap.get(events.get(i)));
+      } else if (i < (events.size() / 2)) {
+        assertEquals(413, resultMap.get(events.get(i)));
```
qingzhuozhen, line 101: The reason here is for add reduce for 413?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -304,40 +316,6 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
     return eventsToRetry;
   }
 
-  private List<Event> pruneEvent(List<Event> events) {
```
qingzhuozhen, line 307: @yuhao900914 Can you confirm this is safe to remove?

yuhao900914: will take a look

yuhao900914: This function is for if we already have <user, device>  pair in the buffer, we directly put the event in the current retry.  
If we take the pruning method out, it will have some unnecessary compute in onEventsError later.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -227,6 +230,7 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
         break;
       case PAYLOAD_TOO_LARGE:
         shouldRetry = true;
+        shouldReduceEventCount = true;
```
yuhao900914, line 233: Do we need to reduceEventCount if the payload is too large?

src/test/java/com/amplitude/HttpTransportTest.java
bohan-amplitude: Yes, since I add reduce payload to operation for status 413. The first rate limit response cut and drop half of the list makes the last 5 callbacks have code 429. The second payload too large response cut and drop 3 events out of 5.
The rest 2 events get success response and callback code is 200.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: This function cause two problems:
1. During the following process, events may be dropped based on invalid response body. These response contains index of events for original list. Events in pruned list got dropped by original list index cause index out of bound exception and we may drop wrong events.
2. During the following process, events may be dropped based on rate limit response body and event's userId/deviceId. The pruned list function skip this check and add events need to drop directly to retry buffer.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: I saw in node sdk. It's set to true.https://github.com/amplitude/Amplitude-Node/blob/main/packages/node/src/retry/defaultRetry.ts
And since response indicate payload is too large. Retry without reduct events make no sense.

qingzhuozhen (approved):

PR #56, opened 2022-02-02
fix: Httptransport concurrency issue

bohan-amplitude:
### Summary

1. Using thread pool to execute retry jobs
2. Using object lock to manage concurrency access of retry buffer and buffer counter
3. change eventOnLoop method to a runnable class

### Checklist


src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -109,100 +111,30 @@ private void onEventsError(List<Event> events, Response response) {
...
+    }
+    for (String userId : users) {
+      Set<String> devices;
+      synchronized (bufferLock) {
+        devices = new HashSet<>(idToBuffer.get(userId).keySet());
+      }
+      for (String deviceId : devices) {
+        try {
+          retryThreadPool.submit(new RetryEventsOnLoop(userId, deviceId));
+        } catch (RejectedExecutionException e) {
```
qingzhuozhen, line 133: What's the possible cases for this exception? And if it is rejected, do we have any further handling?

qingzhuozhen (commented):
Seems this one include the changes in another PR?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -5,11 +5,7 @@
 import org.json.JSONObject;
 
 import java.util.*;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.*;
```
yuhao900914, line 8: Dose import java.util.* include java.util.concurrent.*?

bohan-amplitude: No, they are two packages. import java.util.* won't import classes in sub-packages.

src/main/java/com/amplitude/HttpTransport.java
bohan-amplitude: New tasks submitted in method execute(java.lang.Runnable) will be rejected when the Executor has been shut down, and also when the Executor uses finite bounds for both maximum threads and work queue capacity, and is saturated. For simple, two reason: 1. executor was shutdown. 2 task queue was full if there is a capacity limits. I think we don't need to do further handling since we don't shutdown the executor. New tasks is hard to be refused since we have retry buffer limits.

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -75,9 +74,12 @@ public void sendEventsWithRetry(List<Event> events) {
 
   // The main entrance for the retry logic.
   public void retryEvents(List<Event> events, Response response) {
-    List<Event> eventsToSend = pruneEvent(events);
-    if (eventsInRetry.intValue() < Constants.MAX_CACHED_EVENTS) {
-      onEventsError(eventsToSend, response);
+    int bufferSize;
+    synchronized (counterLock) {
+      bufferSize = eventsInRetry;
```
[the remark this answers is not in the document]
bohan-amplitude: Actually the AtomicInteger is good to use here. I changed to a lock when I debug why buffer size is always negative. I found the problem is not do increment in addEventToBuffer(). But didn't change it back.

qingzhuozhen (approved):
looks good, please make sure all the tests are not affected.

PR #63, opened 2022-04-05
feat: add support for custom headers

justin-fiedler:
### Summary
Added support for setting custom headers. 

Note: I added it to the Options object to match Amplitude-JS, but maybe it would be better to move the headers in Amplitude client instead. Let me know what you think. Thanks!

```
amplitude.setOptions(new Options()
        .addHeader("Custom Header", "value")
        .setMinIdLength(5));
```

### Checklist


src/test/java/com/amplitude/HttpCallTest.java
```diff
@@ -238,4 +270,16 @@ private void verifyConnectionOption(HttpsURLConnection connection) throws Protoc
     verify(connection, times(1)).setReadTimeout(Constants.NETWORK_TIMEOUT_MILLIS);
     verify(connection, times(1)).setDoOutput(true);
   }
+
+  private JSONObject getMockResponse(
```
qingzhuozhen, line 274: nit: this could be `getMockSuccessResponse`

src/demo/java/com/demo/amplitude/DemoController.java
```diff
@@ -1,13 +1,9 @@
 package com.demo.amplitude;
 
+import com.amplitude.*;
```
qingzhuozhen, line 3: Nit: Maybe we can keep the explicit imports. Will create a ticket to add a code style check for Java SDK later.

qingzhuozhen: Oh, this is in demo, I guess either is fine then.

qingzhuozhen (approved):

src/test/java/com/amplitude/HttpCallTest.java
justin-fiedler: I was thinking that too but decided to make the status code configurable for possible reuse in other tests.

PR #72, opened 2022-07-13
feat: Add timeout and cleanUp to Amplitude client

bohan-amplitude:
### Summary

Add more logs when catching exception
trigger call backs when catching exception
Add setFlushTimeout method
Add shutdown interface to close client and release resources

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -279,11 +279,11 @@ public class Amplitude {
   }
 
   /**
-   * Check the status of the client, return true if any of the following situation: 1. Events sit in
-   * memory waiting for flush are more than flush threshold 2. Events in retry buffer are more than
-   * 16,000 3. When setRecordThrottledId(true), the userId or deviceId of the input event was
-   * throttled and retry attempt for the userId/deviceId not finished. Can be called before
-   * logEvent(event) and add some wait time if return true.
+   * Check the status of the client, return true if any of the following situation:
+   * 1. Events sit in memory waiting for flush are more than flush threshold
+   * 2. Events in retry buffer are more than 16,000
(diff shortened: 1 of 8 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -280,6 +293,15 @@ public void setRecordThrottledId(boolean record) {
     httpTransport.setRecordThrottledId(record);
   }
 
+  /**
+   * Release resource by:
+   * Terminate all threads for flushing events
+   * All events hold by these threads will trigger callbacks if amplitude client have configured callbacks method
+   */
+  public synchronized void cleanUp() throws InterruptedException {
```
justin-fiedler, line 301: Should we name this "shudown" to match Ampli SDK?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -105,11 +145,10 @@ private CompletableFuture<Response> sendEvents(List<Event> events) {
           Response response = null;
           try {
             response = httpCall.makeRequest(events);
-            logger.debug("SEND", events, response);
+            logger.debug("SEND", "Events count " + events.size());
+            logger.debug("RESPONSE", response.toString());
           } catch (AmplitudeInvalidAPIKeyException e) {
             throw new CompletionException(e);
-          } catch (Exception e) {
```
liuyang1520, line 111: Just to double check, any reason that we want to remove this safe catch?

bohan-amplitude: moved to outer layer that call sendEvents() so I can handle unexpected exceptions in one place.

liuyang1520 (approved):
Except the minor doc style, other changes LGTM!

PR #74, opened 2022-07-18
fix: fix failed unit tests due to async methods

bohan-amplitude:
### Summary

fix failed unit tests due to async methods

### Checklist


src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -466,4 +469,18 @@ private AmplitudeCallbacks getCallbacks(Amplitude amplitude)
     callbackField.setAccessible(true);
     return (AmplitudeCallbacks) callbackField.get(httpTransport);
   }
+
+  public boolean shouldWaitResultWithTimeout(Amplitude client, Event event, long time, TimeUnit unit) {
```
liuyang1520, line 473: Minor, cannot find any other usage of this helper outside this class, should it be `private` instead?

qingzhuozhen (approved):

PR #76, opened 2022-08-18
feat: add library context field

liuyang1520:
### Summary
- feat: add library context field

Add this option in the configuration, which can be used in the library wrapper/code generation/dynamic loading cases, for setting the `library_context` information.

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/Amplitude.java
+++ b/src/main/java/com/amplitude/Amplitude.java
@@ -24,6 +24,7 @@ public class Amplitude {
   private Object eventQueueLock = new Object();
   private Plan plan;
   private long flushTimeout;
+  private String libraryContext;
 
   /**
    * A dictionary of key-value pairs that represent additional instructions for server save
@@ -164,6 +165,16 @@ public class Amplitude {
   }
 
   /**
+   * Set library context information.
+   *
+   * @param libraryContext String
+   */
+  public Amplitude setLibraryContext(String libraryContext) {
+    this.libraryContext = libraryContext;
+    return this;
+  }
+
+  /**
    * Set custom proxy for https requests
    *
    * @param proxy Proxy object, default to Proxy.NO_PROXY for direct connection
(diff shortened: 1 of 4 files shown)


src/main/java/com/amplitude/Amplitude.java
```diff
@@ -24,6 +24,7 @@ public class Amplitude {
   private Object eventQueueLock = new Object();
   private Plan plan;
   private long flushTimeout;
+  private String libraryContext;
```
liuyang1520, line 27: I found that adding a new middleware to this repo might be a little overkilled, e.g.:
```java
public class ContextMiddleware implements Middleware {
    public String libraryContext;

    public ContextMiddleware(String libraryContext) {
        this.libraryContext = libraryContext;
    }

    @Override
    public void run(MiddlewarePayload payload, MiddlewareNext next) {
        if (payload.event != null) {
            payload.event.libraryContext = this.libraryContext;
        }
        next.run(payload);
    }
}
```

haoliu-amp: What'll be your suggestion?

liuyang1520: Oh, my preference would be the change in this PR.
- The property can be globally set (like `plan`), doesn't need to be applied per event
- Lightweight and minimum changes, adding a new middleware and exporting it in package only for this case seems overkill to me

For the Python (new generation SDk) change: https://github.com/amplitude/Amplitude-Python/pull/33/files. Similarly, exposing a way to set it globally, as Python puts the `library` in ContextPlugin, we can put in the similar place, while keeping the changes minimum and lightweight.

haoliu-amp: Got it! Thanks

qingzhuozhen (approved):

bohan-amplitude (approved):

justin-fiedler (approved):

PR #85, opened 2023-05-25
fix: handle non-json response bodies and retry on 5xx errors.

bgiori:
### Summary

* Catch JSONException thrown by parsing response body & return response with code.
* Consider 5xx errors (Status.FAILURE) for retries.
* Fix `gradlew` by adding the jar file and update the test action to use the wrapper

### Checklist

* [X] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-Java/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?


.github/workflows/semantic-pr.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   pr-title-check:
     name: Check PR for semantic title
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
qingzhuozhen, line 10: thanks for updating this

src/test/java/com/amplitude/HttpTransportTest.java
```diff
@@ -94,9 +100,9 @@ public void onLogEventServerResponse(Event event, int status, String message) {
     httpTransport.setHttpCall(httpCall);
     httpTransport.setCallbacks(callbacks);
     httpTransport.retryEvents(events, invalidResponse);
-    assertTrue(latch.await(1L, TimeUnit.SECONDS));
+    assertTrue(latch.await(1000000L, TimeUnit.SECONDS));
```
qingzhuozhen, line 103: do we need the large number like this?

bgiori: Whoops, thanks for catching this, I was testing :)

Mercy811 (approved):
Hi @bgiori, I am not very familiar with Java SDK. Only had a rough review and want to unblock you.

justin-fiedler (approved):
LGTM. Thanks for the fix @bgiori!

qingzhuozhen (approved):

src/main/java/com/amplitude/HttpCall.java
```diff
@@ -89,6 +90,12 @@ protected Response makeRequest(List<Event> events) throws AmplitudeInvalidAPIKey
       timesOutResponse.put("status", Status.TIMEOUT);
       timesOutResponse.put("code", 408);
       responseBody = Response.populateResponse(timesOutResponse);
+    } catch (JSONException e) {
+      // Some error responses from load balancers and reverse proxies may have
+      // response bodies that are not JSON (e.g. HTML, XML).
+      JSONObject decodeFailureResponse = new JSONObject();
+      decodeFailureResponse.put("code", responseCode);
+      responseBody = Response.populateResponse(decodeFailureResponse);
```
yuhao900914, line 98: Question: we only put the responseCode in responseBody here. Should we try to decode it since we might know it's type will be HTML or XML?

bgiori: Response code is accessed from the http response (protocol) rather than from the actual response body. We don't necessarily care what is in the response body in if it's not json, just the code.

bgiori: The JSON object here wrapping the code is used to conform to what the application protocol expects, which is a `JSONObject` with the response code included.

PR #91, opened 2023-06-23
fix: Should retry 429

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-78787

429 response should indicate that we need to slow down and should retry

### Checklist

### DIFF
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -176,7 +176,6 @@ class HttpTransport {
         break;
       case RATELIMIT:
         if (response.isUserOrDeviceExceedQuote(userId, deviceId)) {
-          shouldRetry = false;
           triggerEventCallbacks(events, response.code, response.error);
         }
         // Reduce the payload to reduce risk of throttling
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -176,7 +176,7 @@ class HttpTransport {
         break;
       case RATELIMIT:
         if (response.isUserOrDeviceExceedQuote(userId, deviceId)) {
-          shouldRetry =false;
+          shouldRetry = false;
           triggerEventCallbacks(events, response.code, response.error);
         }
         break;
--- a/src/main/java/com/amplitude/HttpTransport.java
+++ b/src/main/java/com/amplitude/HttpTransport.java
@@ -176,10 +176,9 @@ class HttpTransport {
         break;
       case RATELIMIT:
         if (response.isUserOrDeviceExceedQuote(userId, deviceId)) {
+          shouldRetry =false;
           triggerEventCallbacks(events, response.code, response.error);
         }
-        // Reduce the payload to reduce risk of throttling
-        shouldReduceEventCount = true;
         break;
       case PAYLOAD_TOO_LARGE:
         shouldRetry = true;
--- a/src/test/java/com/amplitude/AmplitudeTest.java
+++ b/src/test/java/com/amplitude/AmplitudeTest.java
@@ -353,7 +353,6 @@ public class AmplitudeTest {
     HttpCall httpCall = getMockHttpCall(amplitude, false);
     CyclicBarrier barrier = new CyclicBarrier(2);
     CountDownLatch latch = new CountDownLatch(1);
-    CountDownLatch callbackLatch = new CountDownLatch(2);
     Response response = new Response();
     response.status = Status.RATELIMIT;
     response.code = 429;
@@ -382,13 +381,6 @@ public class AmplitudeTest {
     assertFalse(amplitude.shouldWait(event));
     amplitude.setRecordThrottledId(true);
     assertFalse(amplitude.shouldWait(event));
-    amplitude.setCallbacks(
-        new AmplitudeCallbacks() {
-          @Override
-          public void onLogEventServerResponse(Event event, int status, String message) {
-            callbackLatch.countDown();
-          }
-        });
     amplitude.logEvent(event);
     amplitude.logEvent(event2);
     amplitude.flushEvents();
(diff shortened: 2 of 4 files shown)


Mercy811:
I Refer to 429 retry logic from [Python](https://github.com/amplitude/Amplitude-Python/blob/0c8c6a42b983cb5882459df3c1b54fa2bb26c09f/src/amplitude/processor.py#L45), [Node](https://github.com/amplitude/Amplitude-Node/blob/2ef295e1fb698286d606ea4a2ccbbfdc4ba3fdc8/packages/node/src/retry/defaultRetry.ts#L138), and [Go](https://github.com/amplitude/analytics-go/blob/c12a3b0f0af57e7b069a55cab0cb26e9455de5b9/amplitude/plugins/destination/internal/amplitude_response_processor.go#L172)

But not sure how to add a test for it. Any ideas?

src/main/java/com/amplitude/HttpTransport.java
```diff
@@ -176,7 +176,6 @@ private EventsRetryResult retryEventsOnce(String userId, String deviceId, List<E
         break;
       case RATELIMIT:
         if (response.isUserOrDeviceExceedQuote(userId, deviceId)) {
```
justin-fiedler, line 178: We should retry differently based on the response. See [Kotlin implementation](https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/utilities/InMemoryResponseHandler.kt#L66-L71) here and [comments in Slack](https://amplitude.slack.com/archives/C02E1H87E2C/p1687465582561709?thread_ts=1687465015.015709&cid=C02E1H87E2C)

https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/utilities/InMemoryResponseHandler.kt#L66-L71

Mercy811: We don’t have a `eventsToRetryLater` logic available. Can we just delete `shouldReduceEventCount` and retry if don't exceed quote?

justin-fiedler:
> But not sure how to add a test for it. Any ideas?

There are existing texts for 429s. You can add a new one for different mock responses.

https://github.com/amplitude/Amplitude-Java/blob/main/src/test/java/com/amplitude/HttpTransportTest.java#L53

src/main/java/com/amplitude/HttpTransport.java
justin-fiedler: I think [Brian's suggestion seems good](https://amplitude.slack.com/archives/C02E1H87E2C/p1687469723892789?thread_ts=1687465015.015709&cid=C02E1H87E2C). We would still need to test.

> maybe can we just add back the dropped events to the `idToBuffer` map if we drop them due to a 429?

bgiori: I dont think adding back to the idToBuffer map will actually help since that buffer is not tied to the actual buffer in the client.

src/test/java/com/amplitude/HttpTransportTest.java
```diff
@@ -104,12 +104,10 @@ public void onLogEventServerResponse(Event event, int status, String message) {
...
     verify(httpCall, times(5)).makeRequest(anyList());
     for (int i = 0; i < events.size(); i++) {
-      if (i < (events.size() / 4)) {
+      if (i < (events.size() / 2)) {
         assertEquals(200, resultMap.get(events.get(i)));
-      } else if (i < (events.size() / 2)) {
-        assertEquals(413, resultMap.get(events.get(i)));
       } else {
-        assertEquals(429, resultMap.get(events.get(i)));
+          assertEquals(413, resultMap.get(events.get(i)));
```
bgiori, line 110: Fix indentation.

bgiori (approved):

src/test/java/com/amplitude/HttpTransportTest.java
qingzhuozhen, line 110: nit: extra space

src/test/java/com/amplitude/HttpTransportTest.java
```diff
@@ -260,6 +258,47 @@ public void onLogEventServerResponse(Event event, int status, String message) {
       assertEquals(429, resultMap.get(event));
     }
   }
+  
+  @Test
+  public void testRetryEventWithUserThrottled()
```
qingzhuozhen, line 263: Thanks for adding this, do we still plan to add a test case for exploring all retries on rate limit?

src/test/java/com/amplitude/AmplitudeTest.java
```diff
@@ -382,22 +381,14 @@ public void testShouldWait()
...
     amplitude.logEvent(event);
     amplitude.logEvent(event2);
     amplitude.flushEvents();
     assertTrue(shouldWaitResultWithTimeout(amplitude, event, 1L, TimeUnit.SECONDS));
     assertFalse(amplitude.shouldWait(event2));
     barrier.await();
     assertTrue(latch.await(1L, TimeUnit.SECONDS));
-    assertTrue(callbackLatch.await(1L, TimeUnit.SECONDS));
-    assertFalse(shouldWaitResultWithTimeout(amplitude, event, 1L, TimeUnit.SECONDS));
+    assertTrue(shouldWaitResultWithTimeout(amplitude, event, 1L, TimeUnit.SECONDS));
```
qingzhuozhen, line 391: Maybe add a comment on why we flip on this?

src/test/java/com/amplitude/HttpTransportTest.java
justin-fiedler: +1

justin-fiedler (commented):
Thanks @Mercy811 overall looks good. Thanks for the tests. See Qingzhuo's comment about adding new test for max retries on rate limit.

src/test/java/com/amplitude/HttpTransportTest.java
Mercy811: fixed

Mercy811: fixed

src/test/java/com/amplitude/AmplitudeTest.java
Mercy811: This is flipped because when it is a throttled rate limit response, it will retry until max retry times

justin-fiedler (approved):

